PR #4, opened 2017-03-04
Rumor titles shouldn't be crazy long

quad:
So let's make them short.

I originally used punctuation to shorten these up; but, too many rumors have too many crazy uses of punctuation. This seems like a decent compromise.

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -56,6 +56,13 @@ export default compose(
   handleOrderByChange,
   handleFilterChange,
 }) {
+  const shorten_title = (title) => {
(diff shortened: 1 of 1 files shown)


MrOrz:
@quad  Nice!

You are a humorous guy :)

Would you mind changing variable names to camel case and moving the definition of `MAXIMUM_LENGTH` and `shorten_title()`to outside of `Index()`? It is the pretty high up in the virtual dom tree and its render function will be invoked quite a lot. Since it does not use variables in `Index()`'s scope, we does not need to re-declare it every time.

PR #26, opened 2017-08-09
Refactor components and upgrade deps

MrOrz:
1. Broke down large page components into manageable pieces
2. Upgrade next.js to 3.0.1
3. Adhere to common practice of using `connect()`: put it to bottom and separate `mapStateToProps`
4. When `connect()` wraps a functional component and it dispatches actions, use `mapDispatchToProps` to wrap redux action creators with methods that looks like a event handler (`onXXX`). On the other hand, when `connect()` wraps a React class, just handle events right inside the component with `this.props.dispatch`.


util/text.js
```diff
@@ -0,0 +1,34 @@
...
+        s.match(urlRegExp)
+          ? <a key={`link${i}`} href={s}>{shortenUrl(s, maxLength)}</a>
+          : s
+    );
+}
+
+export function nl2br(text = '') {
+  const sentences = text.split('\n');
+  if (sentences.length <= 1) return sentences.map(s => linkify(s));
+  return sentences.map((sentence, i) =>
```
LucienLee: how about this?
```
return str.split(newlineRegex).map(function(line, index) {
    if (line.match(newlineRegex)) {
      return React.createElement('br', { key: index });
    }
    return line;
  });
```

components/ReplyForm.js
```diff
@@ -0,0 +1,202 @@
...
+      reference: localStorage.reference || reference,
+      text: localStorage.text || text,
+    });
+  }
+
+  set(key, value) {
+    this.setState({ [key]: value });
+
+    // Backup to localStorage
+    if (typeof requestAnimationFrame !== 'undefined') {
```
LucienLee: remove check requestAnimationFrame undefined

util/text.js
```diff
@@ -0,0 +1,34 @@
...
+    .map(
+      (s, i) =>
+        s.match(urlRegExp)
+          ? <a key={`link${i}`} href={s}>{shortenUrl(s, maxLength)}</a>
+          : s
+    );
+}
+
+export function nl2br(text = '') {
+  const sentences = text.split('\n');
```
MrOrz: 好喔我換成 non-capturing group
因為 split 如果放 capturing group 的話會多把斷行本身切進去

MrOrz: 喔喔我看懂了，這要搭配你下面的 createElement

util/text.js
MrOrz: 改好了，我把 React.createElement 換成 JSX 惹

PR #37, opened 2017-09-16
Add tooltip and simple similarity on related articles.

ben196888:
- Add tooltip when the cursor hovers related articles.
- Add simple similarity in the header.


MrOrz:
Deployed to production :)

PR #48, opened 2017-10-22
Show content of the related article in related reply section

MrOrz:
When I was processing https://cofacts.g0v.tw/article/AV85NPHeyCdS-nWhuhya
there are a lot of related articles. Later I found an adequate reply inside. If the related article were to shown, it would help me identifying the related reply a lot.

This PR adds the article content above the reply, hoping that it provides good preview for the editors to better choose from existing replies.

### DIFF
--- a/components/RelatedReplies.js
+++ b/components/RelatedReplies.js
@@ -4,6 +4,7 @@ import moment from 'moment';
 import ExpandableText from './ExpandableText';
 import { linkify, nl2br } from '../util/text';
 import { Link } from '../routes';
+import { sectionStyle } from './ReplyConnection.styles';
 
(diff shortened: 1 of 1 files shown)


hazelwei:
Thanks for considering design and I think it would help editors to have a preview of the reply.
One quick question, would all reply show on the list if there are multiple replies list on the related article?

MrOrz:
> would all reply show on the list if there are multiple replies list on the related article?

Yes. If there are 2 replies for "Article A", the related reply list would read:

```
===
Article A
Reply 1 for Article A
---
Article A (again)
Reply 2 for Article A
===
```

PR #51, opened 2017-10-29
Refactor article detail

MrOrz:
Logic revamp:

1. Adding snapshot testing for `articleDetail` reducers. See snapshot files to check how the reducer work!
2. Simplify the logic of processing `relatedReplies`. Previously it inserts `articleId` twice and looking up article in `render()` functions. Now it is more straightforward, with `article` directly embedded.
3. Change related UI components. Removes unused props.

No UI change for this PR.

### DIFF
--- a/components/RelatedReplies.js
+++ b/components/RelatedReplies.js
@@ -88,7 +88,6 @@ function RelatedReplyItem({
 
 export default function RelatedReplies({
   relatedReplies,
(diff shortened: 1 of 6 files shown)


MrOrz:
coverage lowered because we added `redux/articleDetail` to test, but didn't prepare tests for redux actions...

PR #54, opened 2017-11-02
SEO improve

### LINKED ISSUE #52: Change title / subtitle for SEO
As discussed in http://beta.hackfoldr.org/cofacts/https%253A%252F%252Fhackmd.io%252Fs%252FB15v54U0Z

Using "真的假的" in google basically cannot find this project.

We should try the following SEO techniques in our homepage:

- Add 真的假的 to page title
- Add subtitle 
- etc

GoreStarry:
Try to improve SEO by head title and some HTML tags
close #52

### DIFF
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -7,8 +7,10 @@ export default function ArticleItem({ article }) {
   return (
     <Link route="article" params={{ id: article.get('id') }}>
       <a className="item">
-        <div className="item-text">{article.get('text')}</div>
-        <ArticleInfo article={article} />
+        <li>
+          <div className="item-text">{article.get('text')}</div>
+          <ArticleInfo article={article} />
+        </li>
 
         <style jsx>{listItemStyle}</style>
       </a>
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -9,9 +9,9 @@ export default function ArticleItem({ article }) {
       <li className="item">
         <a>
           <div className="item-text">{article.get('text')}</div>
-          <ArticleInfo article={article} /> 
+          <ArticleInfo article={article} />
         </a>
-        
+
         <style jsx>{listItemStyle}</style>
       </li>
     </Link>
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -6,14 +6,14 @@ import { listItemStyle } from './ListItem.styles';
 export default function ArticleItem({ article }) {
   return (
     <Link route="article" params={{ id: article.get('id') }}>
-      <a className="item">
-        <li>
+      <li className="item">
+        <a>
           <div className="item-text">{article.get('text')}</div>
-          <ArticleInfo article={article} />
-        </li>
-
+          <ArticleInfo article={article} /> 
+        </a>
+        
         <style jsx>{listItemStyle}</style>
-      </a>
+      </li>
     </Link>
   );
 }
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -5,15 +5,15 @@ import { listItemStyle } from './ListItem.styles';
 
 export default function ArticleItem({ article }) {
   return (
-    <Link route="article" params={{ id: article.get('id') }}>
-      <li className="item">
+    <li className="item">
+      <Link route="article" params={{ id: article.get('id') }}>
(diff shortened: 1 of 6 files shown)


components/ArticleItem.js
```diff
@@ -7,8 +7,10 @@ export default function ArticleItem({ article }) {
   return (
     <Link route="article" params={{ id: article.get('id') }}>
       <a className="item">
```
MrOrz: The `ul` element [does not support a tag](http://w3c.github.io/html/single-page.html#the-ul-element) as its direct children. How about swapping `a` and `li` to make the structure with the spec, and alter the CSS so that it looks the same?

pages/article.js
```diff
@@ -209,11 +209,12 @@ class ArticlePage extends React.Component {
     }
 
     const structuredData = this.getStructuredData();
+    const slicedArticleTitle = article.get('text').slice(0, 15);
 
     return (
       <div className="root">
         <Head>
-          <title>{article.get('text').slice(0, 15)}⋯⋯ - 文章</title>
+          <title>{slicedArticleTitle}⋯⋯ | 真的假的</title>
```
MrOrz: I think we can make it `| Cofacts 真的假的`  so that both EN and ZH names goes into the title

pages/index.js
```diff
@@ -120,9 +121,9 @@ class Index extends ListPage {
     return (
       <main>
         <Head>
-          <title>文章列表</title>
+          <title>真的假的 - 轉傳訊息查證</title>
```
MrOrz: Let's make it `Cofacts 真的假的 - 轉傳訊息查證`

components/ArticleItem.js
```diff
@@ -6,12 +6,14 @@ import { listItemStyle } from './ListItem.styles';
 export default function ArticleItem({ article }) {
   return (
     <Link route="article" params={{ id: article.get('id') }}>
-      <a className="item">
-        <div className="item-text">{article.get('text')}</div>
-        <ArticleInfo article={article} />
+      <li className="item">
+        <a>
```
MrOrz: `<a>` tag must be the direct children of `<Link>` so that `<Link>` can properly configure `href` property.

Please refer to the image below:

## master branch  - `<Link>` is the direct parent of `<a>`

Note that `<a>` contains `href` here, so that search engines can pick up the links easily.


## This PR - No href in `<a>` tag

Although the links are still clickable, it was [handled by onClick event within next.js](https://github.com/zeit/next.js/#using-a-component-that-support-onclick). Search engines may not be able to pick the links up properly.




In order to fix this, we should use the following structure, so that
* `<li>` is `<ul>`'s direct child
* `<a>` is `<Link>`'s direct child
* Only 1 child is allowed in `<Link>`

```js
    <li className="item">
      <Link route="article" params={{ id: article.get('id') }}>
        <a>
          <div className="item-text">{article.get('text')}</div>
          <ArticleInfo article={article} />
        </a>
      </Link>
      <style jsx>{listItemStyle}</style>
    </li>
```

Note that in this way, we should also `listItemStyle` must be edited to remove the underline of hyperlinks.

pages/index.js
```diff
@@ -120,9 +121,9 @@ class Index extends ListPage {
     return (
       <main>
         <Head>
-          <title>文章列表</title>
+          <title>Cofacts 真的假的 - 轉傳訊息查證</title>
         </Head>
-
+        <h2>文章列表</h2>
```
MrOrz: How about adding `<h2>回應列表</h2>` to `page/replies.js`, since the reply list page should be symmetric to article list page?

components/ArticleItem.js
GoreStarry: make sense, Great!

pages/article.js
```diff
@@ -227,14 +228,14 @@ class ArticlePage extends React.Component {
 
         <section className="section">
           <header className="header">
-            <h2>訊息原文</h2>
+            <h2>訊息原文 - {slicedArticleTitle}...</h2>
```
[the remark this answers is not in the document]
GoreStarry: I think h2 is more like article title, but since the sliced title actually not the proper title, so maybe it does not appropriate. I will recover it.
(especially SEO for article page is enough at the beginning.)

PR #55, opened 2017-11-12
Upgrade NodeJS and deps

MrOrz:
Note: Everyone should run `npm install` after this change!

* **Drops yarn**; use npm 5
* Upgrades to React 16, Next 4
* Updates Dockerfile
* Update all coding styles due to prettier update


MrOrz:
I'm merging this because it has a large changelog and blocks others from editing

PR #58, opened 2017-11-21
14 reply feedback

### LINKED ISSUE #14: 讓使用者可以針對回應評價「你覺得這個回應有用嗎」
讓網站使用者可以登入之後送出「你覺得這個回應有用嗎」（`ReplyFeedback`）

要使用的 API：`CreateOrUpdateReplyConnectionFeedback` mutation
Doc: https://cofacts-api.hacktabl.org/graphql

GoreStarry:
Add buttons for reply feedback.
* voteReply action creator
* move feedback status & buttons to an Individual component(ReplyFeedback)
* get user auth id pass to ReplyFeedback component to check if reply belongs to user itself

Close #14


components/ReplyFeedback.styles.js
```diff
@@ -0,0 +1,58 @@
+import css from 'styled-jsx/css'; // eslint-disable-line import/no-unresolved
+
+export const feedbackStyle = css`
+  .ReplyFeedback {
```
MrOrz: Let's use `dashed-case` for all class selectors :D

PR #69, opened 2017-12-09
ArticleItem widget for pc and mobile

### LINKED ISSUE #65: Index ArticleItem "HAVE READ" / "NOT ARTICLE" quick control widget.
Index ArticleItem "HAVE READ" / "NOT ARTICLE" quick control widget.

GoreStarry:
ArticleItem widget for pc and mobile
close #65


GoreStarry:
嗯travis過不了... 我沒有改articleDetail.js這隻啊... 我不知道發生了什麼事情...

pages/index.js
```diff
@@ -17,11 +17,57 @@ import { load, loadAuthFields } from 'ducks/articleList';
...
+      );
+      localEditorHelperList &&
+        this.setState({
+          localEditorHelperList,
+        });
+    }
+  };
+
+  handleLocalEditorHelperList = ({ id, read, notArticleReplied }) => {
+    this.setState(
```
MrOrz: Since we are reading and setting `localEditorHelperList` at the same time, we should use the updater function style instead:

```
this.setState(({localEditorHelperList}) => ({
  ...localEditorHelperList,
  [id]: {read, notArticleReplied}
}), cb);
```

Ref: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

pages/index.js
```diff
@@ -130,19 +176,27 @@ class Index extends ListPage {
...
-          {articles.map(article => (
-            <ArticleItem
-              key={article.get('id')}
-              article={article}
-              requestedForReply={authFields.get(article.get('id'))}
-            />
-          ))}
+          {articles.map(article => {
+            const id = article.get('id');
+            const EditorHelperState = localEditorHelperList[id];
```
MrOrz: Since it's not a class, let's keep it lower-cased.
Or maybe just `{...localEditorHelperList[id]}` in line 196, so that we don't need a variable at all

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,255 @@
...
+
+export default class ArticleItemWidget extends PureComponent {
+  static propTypes = {
+    id: PropTypes.string.isRequired, // Article ID
+    read: PropTypes.bool.isRequired,
+    notArticleReplied: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.string, // TYPE_SUGGESTION_OPTIONS in replyType.js
+    ]),
+    handleLocalEditorHelperList: PropTypes.func.isRequired,
```
MrOrz: Let's keep `handleXXX` to the actual "handler" that implements the change.

`ArticleItemWidget` is invoking the callback "on" some event, so we should use `onXXX` to name the handlers here.

I think `onLocalEditorHelperListChange` would be a bit too long, `onChange` would be appropriate for this case since `ArticleItemWidget` could only "change" in one way.

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,255 @@
...
+      barOpen: true,
+      readHover: true, //for mobile, explain in onTouchMoveTip function
+      touchTarget: 'read',
+    });
+  };
+
+  hoverTag = event => {
+    const name = event.currentTarget.attributes['name'].value;
+    this.setState({
+      [`${name}Hover`]: true,
```
MrOrz: Is it possible to hover multiple items at once?

If not, we should just go for one single "hoverIdx" in state, instead of multiple `xxxHover`s. This way we can get rid of `this.initialTagsState`.

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,255 @@
...
+        ref={widget => (this.refWidget = widget)}
+        style={{ zIndex: notArticleSelectionDisplay ? 2 : 1 }} //or the extend component will cover by next ul
+        onClick={this.preventAll}
+        onMouseEnter={this.openBar}
+        onMouseLeave={this.onMouseLeave}
+        onTouchStart={this.openBar}
+        onTouchMove={this.onTouchMoveTip}
+        onTouchEnd={this.onBarTouchEnd}
+      >
+        {this.tags.map((tag, index) => {
```
MrOrz: Personally I would like more declarative style, like:

```
<ul>
  <WidgetItem
    description="已讀"
    color="#cfcfcf"
    hover={this.state.hoverIdx === 0}
    index={0}
    onClick={this.handleReadClick}
    onMouseEnter={this.hoverTag}
    onMouseLeave={this.unHoverTag}
  />
  <WidgetItem
    description="非查證"
    color="orange"
    hover={this.state.hoverIdx === 1}
    index={1}
    onClick={this.handleNotArticleClick}
    onMouseEnter={this.hoverTag}
    onMouseLeave={this.unHoverTag}
  >
    {this.state.notArticleSelectionDisplay && <NotArticleExtendSelection replyToNotArticle={this.replyToNotArticle} />}
  </WidgetItem>
</ul>
```

So that we don't need `this.tags`, which is a bit imperative.

GoreStarry: I think I use the map function to deal with this only for get the index, then feed it to CSS variable. but obviously, it's not really necessary~

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,242 @@
...
+    var changedTouch = event.changedTouches[0];
+    var elem = document.elementFromPoint(
+      changedTouch.clientX,
+      changedTouch.clientY
+    );
+    try {
+      var target_name = elem.attributes['name'].value;
+      target_name &&
+        this.setState(
+          Object.assign({}, this.initialTagsState, {
```
MrOrz: Is it necessary to reset all other state to initialState?

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,242 @@
...
+      // touch move not on the widget tag
+    }
+  };
+
+  // trigger the onClick function in tags config, when finger touch end with tag elm
+  onBarTouchEnd = event => {
+    this.preventAll(event);
+    const target = this.tags.find(select => {
+      return select.name === this.state.hoverTargetName;
+    });
```
MrOrz: Since we do not have this.tags for now, we should replace this with a switch statement.

```
switch(this.state.hoverTargetName){
  case 'read': return this.handleReadClick();
  case 'notArticle': return this.handleNotArticleClick();
}
```

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,242 @@
...
+      hoverTargetName: false,
+    });
+  };
+
+  /**
+   * since the touch start event can only trigger when the finger down at the first tag,
+   * so it's hard to done with like mouse enter/leave strategy
+   * then I use onTouchMove to dynamic check which tag was touch over
+   */
+  onTouchMoveTip = event => {
```
MrOrz: Let's follow the convention to use `handleXXX` naming for `onTouchMoveTip`, `onBarTouchEnd` and `onMouseLeave`, because they are handling events and are not props.

GoreStarry: So, is that better replace the name of function "hoverTag" with "handleHoverTag"?

components/ArticleItemWidget/ArticleItemWidget.js
GoreStarry: You're right. It's not necessary anymore.

components/ArticleItemWidget/ArticleItemWidget.js
MrOrz: To be precise, it would be `handle + object (noun) + verb`, so it should be `handleTagHover`.

In practice, I didn't mention `hoverTag()` because it's name does not confuse with `onXXX`. When I see `onXXX` I would assume that it's not implemented in the component, but passed down from parents.

components/ArticleItemWidget/ArticleItemWidget.js
```diff
@@ -0,0 +1,242 @@
...
+    var elem = document.elementFromPoint(
+      changedTouch.clientX,
+      changedTouch.clientY
+    );
+    try {
+      var target_name = elem.attributes['name'].value;
+      target_name &&
+        this.setState(
+          Object.assign({}, this.initialTagsState, {
+            [`${target_name}Hover`]: true,
```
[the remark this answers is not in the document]
GoreStarry: It should be abandoned by the previous version, I missed removing this one, thanks to finding it!!

components/ArticleItemWidget/ArticleItemWidget.js
GoreStarry: Actually, I'm not really understanding how to naming function appropriately, so really helpful for your explanation. Thanks! and I will change all about that.

components/ArticleItemWidget/ArticleItemWidget.js
GoreStarry: What a big mistake... thanks~~

PR #76, opened 2018-01-20
46 new replies tabbed

GoreStarry:
* the tab of search 
* list of search articles
* modal for list article replies
* add overflow hidden for body when modal open
* List of search replies and search articles for redux articleDetail


components/ReplySearch/ReplySearch.js
```diff
@@ -0,0 +1,170 @@
...
+
+    return (
+      <div>
+        <label htmlFor="replySeach">
+          搜尋相關回應：
+          <input id="replySeach" type="search" onKeyUp={this.handleSearch} />
+        </label>
+
+        {articles.size || replies.size
+          ? [
```
MrOrz: This array here introduces key errors:




Please use `<Fragment>` instead.

Although we did not upgrade babel to support `<>` and `</>`, we can still import `<Fragment>`:

```
import React, {Fragment} from React;
```

components/ReplySearch/ReplySearch.js
```diff
@@ -0,0 +1,170 @@
...
+      </ul>
+    );
+  };
+
+  renderSearchReplyTab = () => {
+    const { articles, replies, onConnect } = this.props;
+    const { tab } = this.state;
+
+    const getArticleSimilarity = relatedArticleText =>
+      stringSimilarity.compareTwoStrings(this.search, relatedArticleText);
```
MrOrz: `this.search` is undefined. Isn't `search` in `this.state`?

Error screenshot:

ducks/articleDetail.js
```diff
@@ -300,6 +410,37 @@ export default createReducer(
...
+          'node',
+          'replyConnections',
+          '0',
+          'article',
+        ]);
+        const version = reply.getIn(['node', 'versions']);
+        return Map()
+          .setIn(['id'], reply.getIn(['node', 'id']))
+          .setIn(['article'], article)
+          .setIn(['versions'], version);
```
MrOrz: Let's do this to simplify things up:
```
Map({
  id: reply.getIn(['node', 'id']),
  article: reply.getIn(['node', 'replyConnections', 0, 'article']),
  versions: reply.getIn(['node', 'versions']),
})
```

components/ReplySearch/ReplySearch.js
GoreStarry: HAHA... big reconstruct mistake, THX!!

PR #80, opened 2018-02-10
remove overflow hidden and fixed from modal open

GoreStarry:
I set body fixed and overflow hidden when modal open. But it makes search list hard to use, maybe should remove it...


MrOrz:
Thanks for the fix!

PR #103, opened 2018-04-29
Remove reply version from all logic

MrOrz:
This PR removes all deprecated `replies.versions`.

After this we can finally remove the deprecated fields in rumors-api.


MrOrz:
I need this to be merged now. Sorry bro.

PR #115, opened 2018-06-11
Language switch for landing page

MrOrz:
Provides both English and Traditional Chinese version of landing page.

When visiting `/`: Use `Accept-Language` to determine. Default to `en`.
When visiting `/en`: Use `en`
When visiting `/tw`: Use `tw`

The `/en` and `/tw` are meant to implement language override switch. However, the switch itself (can be implemented as simple as an hyperlink to `/en` or `/tw`) is not included in this PR. Patches welcome!


MrOrz:
Deployed to staging.

You can visit https://cofacts.hacktabl.org/ and test different language by changing your browser's language to `en` first or `zh` first.

Issue #129, opened 2018-08-19
UserName causing logout to error


MrOrz:
Reproducing on staging:
1. Login
2. Logout
3. See error
![err](https://user-images.githubusercontent.com/108608/44305418-45c2e080-a3a9-11e8-83b7-911bdd712110.gif)

Reproduction on localhost:
<img width="1227" alt="2018-08-19 12 19 30" src="https://user-images.githubusercontent.com/108608/44305423-730f8e80-a3a9-11e8-8dbd-c136f8fe6bfc.png">

I don't know the root cause. `UserName`'s `componentDidUpdate` looks fine to me.

Why would `this.props.userName` become `null`? It needs further investigation.

MrOrz:
Fixed in #131

PR #131, opened 2018-08-24
Fix level-up detection causing logout to fail

MrOrz:
As specified by #129 ,
Previously logging out may cause error (reproducible on [staging](https://cofacts.hacktabl.org/articles) now).

This PR fixes the issue by handling the case where the new `user` becomes null after logout.

After fix:
logout


MrOrz:
Now deployed to staging

PR #136, opened 2018-08-26
#133: fixed the bug that the createdAt field of a reply is undefined

### LINKED ISSUE #133: The createdAt field of replies is not set
This field is not set when creating replies such that replies are always created "a few seconds ago".
screenshot from 2018-08-25 16-11-01

changhc:
Fixed #133.
Note that here I used the createdAt field in the ArticleReply model (ReplyConnection), i.e. the time a reply is created under an article no matter it's a brand new reply or a quote.


changhc:
...I forgot to update the snapshots orz

Issue #142, opened 2018-10-16
URL preview breaks some article pages


MrOrz:
When visiting this page, it gives 500 internal server error.

https://cofacts.g0v.tw/article/1goyooqk2mtrh

the error log in production server (Nothing in Rollbar...)
```
site_1          | { TypeError: Cannot read property 'size' of null
site_1          |     at Hyperlinks (/srv/www/.next/server/bundles/pages/article.js:2955:18)
site_1          |     at d (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:26:492)
site_1          |     at wa (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:28:493)
site_1          |     at a.render (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:33:48)
site_1          |     at a.read (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:32:288)
site_1          |     at renderToString (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:43:6)
site_1          |     at renderPage (/srv/www/node_modules/next/dist/server/render.js:275:26)
site_1          |     at Function.getInitialProps (/srv/www/.next/server/bundles/pages/_document.js:181:25)
site_1          |     at _callee$ (/srv/www/node_modules/next/dist/lib/utils.js:111:30)
site_1          |     at tryCatch (/srv/www/node_modules/regenerator-runtime/runtime.js:62:40) sourceMapsApplied: true }
```

MrOrz:
looks like article page & reply page is both having issues

```
site_1          | { TypeError: Cannot read property 'size' of null
site_1          |     at Hyperlinks (/srv/www/.next/server/bundles/pages/article.js:2955:18)
site_1          |     at d (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:26:492)
site_1          |     at wa (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:28:493)
site_1          |     at a.render (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:33:48)
site_1          |     at a.read (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:32:288)
site_1          |     at renderToString (/srv/www/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:43:6)
[...]

PR #143, opened 2018-10-16
<Hyperlinks> prop null handling

MrOrz:
# Root cause
API server would return `null` for `hyperlinks` field for some articles.

# Proposed fix
Handle `null` in `<Hyperlinks>`

# Screenshot
Connect localhost to production API server, can open the page that errors in production:


MrOrz:
Thanks for the timely review!

MrOrz:
This should fix #142

PR #156, opened 2019-03-13
Make url preview less wide

agameofprivacy:
Make it less wide by setting max-width css on `link` class.

**Before**
Screen Shot 2019-03-13 at 7 26 47 PM

**After**
Screen Shot 2019-03-13 at 7 25 39 PM


MrOrz:
Thanks a lot!

Issue #120, opened 2018-07-08
Should show reasons why users downvote a reply


MrOrz:
In API, `ArticleReplyFeedback` object type has `comment` field containing the reason why a LINE user downvotes a reply. These reasons should be shown in the website so that editors can improve the reply.

In the reply feedback count display, we should display the list of  `comment` (reasons) why the users think the article-reply is useless.

For minimal UI, we can put a link beside the downvote count and leverage the current `<Modal>` to display the downvote reasons:

<img width="211" alt="2019-03-06 10 26 33" src="https://user-images.githubusercontent.com/108608/53888252-f7341580-405e-11e9-9add-b963f6f86ad4.png">

<img width="253" alt="2019-03-06 10 24 25" src="https://user-images.githubusercontent.com/108608/53888122-acb29900-405e-11e9-92e3-5c9ca51d147c.png">

MrOrz:
Oops, it's a duplicate of #97 . Cloising this.

MrOrz:
Reopen because #97 is now scoped down to submission of reason only.

leannechen:
請問我可以認領這個嗎～

PR #159, opened 2019-04-14
Show reasons why a reply is downvoted

leannechen:
### Issue
https://github.com/cofacts/rumors-site/issues/120

### Feature
Users can view reasons why a reply is downvoted by clicking "Why?" beside downvote count. 

### Screenshot
- Default
cofacts_show-reasons-of-downvote_with-legend

- Downvote Reasons Modal
cofacts_show-reasons-of-downvote-2

- If no downvote, do not show switch
cofacts_show-reasons-of-downvote-3

### DIFF
--- a/components/ReplyFeedback.js
+++ b/components/ReplyFeedback.js
@@ -9,6 +9,9 @@ class ReplyFeedback extends Component {
     replyConnection: PropTypes.object.isRequired,
     onVote: PropTypes.func.isRequired,
   };
+  state = {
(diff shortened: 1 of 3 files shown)


leannechen:
@MrOrz Indeed it's simpler, thanks for the advice!
Fixed in https://github.com/cofacts/rumors-site/pull/159/commits/ea57bd99fe59b82faac83f3299b6445b4c6fd04a

PR #166, opened 2019-09-24
Existing reply list in article page

MrOrz:
## Feature
Article page "existing reply" section, includes:
- Component hierarchy: `[id].js` (article page) --> `<CurrentReplies>` covers all article replies & deleted article replies --> list of `<ArticleReply>` --> `<EditorName>`, `<CopyButton>`, `<ReplyFeedback>` in each `<ArticleReply>`.
- Add `<CurrentReplies>` to `[id].js`.
- Refactor `<CurrentReplies>` to use apollo-client; process mutation of existing replies at this level.
- `ReplyConnection.js` renamed to `ArticleReply.js` and rewritten using apollo-client
- Replace `components/Modal` use with material-ui `Dialog`; Replace `balloon-css` with material-ui `Tooltip`

## Refactor
- adds `useCurrentUser` hook that allows users to check if an user is logged in. Caching is done by apollo client.
- uses `BatchHttpLink` to replace original `HttpLink` so that graphql requests in same render will be gathered in one HTTP request.
- rename `Hyperlinks` fragment variable name to match convention
- handle i18n for date-fns `format()` and `formatDistanceToNow()` in `lib/dateWithLocale.js` - [Reference](https://date-fns.org/docs/I18n)

### Article page

螢幕快照 2019-09-24 下午3 04 04

###  "Why" dialog
螢幕快照 2019-09-24 下午3 04 11

### Deleted article reply dialog
螢幕快照 2019-09-24 下午3 04 51

### Voting replies will refresh reply order
downvote

### Deleting and restoring replies
delete-reply

restore


pages/article/[id].js
```diff
@@ -67,6 +76,8 @@ function ArticlePage({ query }) {
     [data, dataForUser]
   );
 
+  const replySectionRef = useRef(null);
```
Comment (LucienLee) [inline pages/article/[id].js]: what's that ref used for?

Comment (MrOrz) [reply pages/article/[id].js]: Ohh good question. It's not used for this PR.
After an editor submits a new reply, the ref is used to scroll the user to the new reply (top section of the reply).

The logic can be seen here in the old codebase: https://github.com/cofacts/rumors-site/blob/master/pages/article.js#L109-L114

MrOrz:
I found that the snackbar after clicking copy button is not shown. Investigating.

MrOrz:
> I found that the snackbar after clicking copy button is not shown. Investigating.

怪，又出現了，merge ㄅ

PR #168, opened 2019-09-30
Create new reply form

MrOrz:
- Introduce `<NewReplySection>`, which will be the functional component performing apollo hooks for "Create reply" (this PR), "use existing reply" (future PR) and "search" (future PR)
- Component structure:
  - `[id].js` (article page)
    - `NewReplySection.js` the 3-tab functional component doing apollo-hooks
      - `ReplyForm.js` the "new reply" form
- `ReplyForm.js` no longer relies on its `onSubmit` returning a promise to clear the field. Instead, it exposes a public method `clear()` for its parent component to clear when reply submission succeeded. This API is similar to how HTML inputs are cleared.

### Before login
螢幕快照 2019-09-30 下午1 27 40

### After login
螢幕快照 2019-09-30 下午1 28 11

### Automatically scroll & clear form after submission
submission

## Refactor
During development, I found that the `merge()` in `[id].js` is actually causing issues. I tried to use `merge()` to merge `LoadArticlePage` and `LoadArticlePageForUser` operations; however, the cache of these two operations are not guaranteed to be updated in the same time. When `LoadArticlePage` and `LoadArticlePageForUser` has inconsistent order of `articleReplies`, `merge()` will erroneously merge the two arrays, causing issue like duplicated `replyIds`:
螢幕快照 2019-09-29 上午2 06 20

To solve this, we should avoid doing `merge()` by ourselves, let apollo-client to update the normalized cache instead. In this PR, all fields in `xxxForUser` fragments are merged into `xxxData` fragments, so that refetching `LoadArticlePage` will fetch and update the for-user fields (like `ownVote`) as well.

### DIFF
--- a/components/NewReplySection.js
+++ b/components/NewReplySection.js
@@ -1,0 +1,42 @@
+import { useState, useCallback } from 'react';
+import { t } from 'ttag';
+
+import Tabs from '@material-ui/core/Tabs';
+import Tab from '@material-ui/core/Tab';
+import Badge from '@material-ui/core/Badge';
+
+import useCurrentUser from 'lib/useCurrentUser';
+import ReplyForm from './ReplyForm';
+
+function NewReplySection({ onSubmit }) {
+  const [selectedTab, setSelectedTab] = useState(0);
+  const currentUser = useCurrentUser();
+
+  const handleTabChange = useCallback((e, v) => setSelectedTab(v), []);
+  const handleSubmit = useCallback(() => {
+    onSubmit(); // Notify upper component of submission
+  }, [onSubmit]);
+
+  if (!currentUser) {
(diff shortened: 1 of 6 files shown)


components/NewReplySection.js
```diff
@@ -0,0 +1,95 @@
...
+
+  if (!currentUser) {
+    return <p>{t`Please login first.`}</p>;
+  }
+
+  return (
+    <>
+      <Tabs value={selectedTab} onChange={handleTabChange}>
+        <Tab label={t`New Reply`} />
+        <Tab
```
MrOrz: 2nd and 3rd tabs are fake. Will implement in the future PR.

PR #178, opened 2019-11-21
Fix HTTP header & handle empty users

MrOrz:
Deployed to https://cofacts.hacktabl.org/

This PR fixes:
- Correct `x-app-id` should be provided in GraphQL request headers so that current user can be recognized as author of new replies
- Handle the case when empty reply.user causes exceptions

We will be deploying this to production soon. Please try navigating pages and submit replies in https://cofacts.hacktabl.org/ to see if there are any unhandled exceptions, thanks!

### DIFF
--- a/components/ArticleReply.js
+++ b/components/ArticleReply.js
@@ -175,8 +175,8 @@ class ArticleReply extends React.PureComponent {
     const originalAuthorElem = (
       <EditorName
         key="editorName"
-        editorName={replyAuthor.name}
-        editorLevel={replyAuthor.level}
+        editorName={replyAuthor?.name}
+        editorLevel={replyAuthor?.level}
       />
     );
     const originalAuthorsReply = (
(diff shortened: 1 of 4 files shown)


components/ArticleReply.js
```diff
@@ -175,8 +175,8 @@ class ArticleReply extends React.PureComponent {
     const originalAuthorElem = (
       <EditorName
         key="editorName"
-        editorName={replyAuthor.name}
-        editorLevel={replyAuthor.level}
+        editorName={replyAuthor?.name}
```
LucienLee: Does js have optional syntax as typescript does now?

components/ReplyFeedback.js
```diff
@@ -90,7 +90,10 @@ function ReplyFeedback({
     onCompleted: () => setReorderSnackShow(true),
   });
 
-  const isOwnArticleReply = currentUser?.id === user.id;
+  // Note that currentUser and user may be undefined or null (when appId mismatch)
+  // Noth case should not consider as ownArticleReply
```
LucienLee: both?

components/ArticleReply.js
MrOrz: We are using https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
It's currently in stage 3: https://github.com/tc39/proposal-optional-chaining

components/ReplyFeedback.js
MrOrz: thanks lol

PR #179, opened 2019-11-24
Google Tag Manager support

MrOrz:
This PR adds Google Tag Manager support.

- Google Tag Manager embedding snippet
- utility function to push to dataLayer: `lib/gtm.js`
- emits custom events like `routeChangeStart`, `routeChangeComplete`, and `dataLoaded` so that developers can build triggers on that

### DIFF
--- a/lib/gtm.js
+++ b/lib/gtm.js
@@ -1,0 +1,13 @@
+/**
+ * Google tag manager related utility functions
+ */
+
+/**
+ * @param  {...any} args - data to feed to dataLayer.push
+ */
+export function pushToDataLayer(...args) {
+  if (typeof window === 'undefined') return; // Skip in SSR
+
+  if (!window.dataLayer) window.dataLayer = [];
+  window.dataLayer.push(...args);
+}
--- a/lib/gtm.js
+++ b/lib/gtm.js
@@ -15,15 +15,12 @@ export function pushToDataLayer(...args) {
 }
 
 /**
- * @param {object} props - data to send to pushDataLayer on component mount
+ * @param {*} trigger - trigger useEffect() after trigger turns truthy
+ * @param {object} args - data to send to pushDataLayer on trigger change
  */
-export function PushToDataLayer({ ...props }) {
+export function usePushToDataLayer(trigger, args) {
   useEffect(() => {
-    // Make sure all rendered elements has been flushed to DOM
-    setTimeout(() => {
-      pushToDataLayer(props);
-    }, 0);
-  }, []);
-
-  return null;
+    if (!trigger) return;
+    pushToDataLayer(args);
+  }, [!trigger]);
 }
--- a/lib/gtm.js
+++ b/lib/gtm.js
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+
 /**
  * Google tag manager related utility functions
  */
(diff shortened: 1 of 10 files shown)


lib/gtm.js
```diff
@@ -0,0 +1,29 @@
...
+  if (typeof window === 'undefined') return; // Skip in SSR
+
+  if (!window.dataLayer) window.dataLayer = [];
+  window.dataLayer.push(...args);
+}
+
+/**
+ * @param {object} props - data to send to pushDataLayer on component mount
+ */
+export function PushToDataLayer({ ...props }) {
```
LucienLee: Is it a good way to declare react effect function with the same name in capitalization?

lib/gtm.js
```diff
@@ -0,0 +1,29 @@
...
+  useEffect(() => {
+    // Make sure all rendered elements has been flushed to DOM
+    setTimeout(() => {
+      pushToDataLayer(props);
```
LucienLee: hmm...do we really need to wait for a macro task?
That means when component did mount, it doesn't mean that dom updated.

lib/gtm.js
MrOrz: Hmm I invented this, not seeing this pattern somewhere else.

I wrap the function as a functional component because I feel that it blends well with apollo-client's hooks. It will look more complicated if we implement this logic in useEffect hooks because we will need to duplicate the if-else logic regarding data loading flag inside the hook.

lib/gtm.js
MrOrz: Good point. I added this because by the time this is invoked, the title is not flushed to DOM yet (probably because <Head> also works asynchronously). But I agree it does not make sense here.

Maybe the developer should handle timeout in Google Tag Manager instead?

MrOrz:
Forgot to mention that this PR also disables static-page optimization for landing page (`pages/index.js`), as discussed with @LucienLee  in 11/20 meeting.

lib/gtm.js
LucienLee: Yeah, this fp will be helpful, but I'm not sure the naming is good.
how about name it as `usePushToDataLayer`?

lib/gtm.js
LucienLee: For example, `ReactGA` pass all parameter by manually passing props rather than part from page. That's might be a more react way. 

I'm okay with both ways but noted that what we waited is head meta elements

MrOrz: Yeah I agree. But the point of using Google Tag Manager instead of triggering `ga()` in source code is to enable designers and developers to add whatever tracking and track whatever event they want without deploying new versions of source code. That's why I would like to focus on the timing of 'trigger' (which can be integrated to any 'tag') rather than how to get Google analytics event done.

lib/gtm.js
MrOrz: But this is a component, not a hook.

MrOrz: When a React component integrated with apollo-client, it looks like this:

```javascript
function Component() {
  const {data, loading, error} = useQuery(...) ;
  if(loading) {
    return <>...</>
  }
  if(error) {
    return <>...</>
  }

  return <>...</>
}
```

When `PushToDataLayer` is implemented as a component, we can just put it in the last return statement:
```javascript
function Component() {
  const {data, loading, error} = useQuery(...) ;
  if(loading) {
    return <>...</>
  }
  if(error) {
    return <>...</>
  }

  return <>...<PushToDataLayer ...></>
}
```

But if we implement it as hooks, we will need to duplicate the if statement:

```javascript
function Component() {
  const {data, loading, error} = useQuery(...) ;
  usePushToDataLayer(loading, error);

  if(loading) {
    return <>...</>
  }
  if(error) {
    return <>...</>
  }

  return <>...</>
}

function usePushToDataLayer(loading, error) {
  if(loading || error) return;
  useEffect(() => {
    pushToDataLayer(...);
  }, [])
}
```

LucienLee: Both ways are okay for me. 
if using functional component -> I think that naming might be need to changed. 
if using custom effect -> I think that duplicated if-else is not a problem. Because of that if-else give this function more cohesion which is like the concept __AOP__

```js
function usePushToDataLayer(enable) {
  if(!enable) return;
  useEffect(() => {
    pushToDataLayer(...);
  }, [])
}
```

MrOrz: Updated.

PR #188, opened 2019-12-09
Add /analytics & /hack redirect page

MrOrz:
[Previously](https://github.com/cofacts/rumors-site/blob/pocky-day-gathering/server.js), visiting `cofacts.g0v.tw/analytics` redirects to google data studio and `cofacts.g0v.tw/hack` redirects to our hackfoldr.

This PR reimplements such redirection mechanism, in [next.js v9](https://nextjs.org/docs#fetching-data-and-component-lifecycle) way.

analytics-redirect

### DIFF
--- a/pages/analytics.js
+++ b/pages/analytics.js
@@ -1,0 +1,15 @@
+function AnalyticsPage() {
+  return null;
+}
+
+AnalyticsPage.getInitialProps = function({ res }) {
+  // Redirect to analytics
+  res.writeHead(303, {
+    Location:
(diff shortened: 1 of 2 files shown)


MrOrz:
Actually `/analytics` and `/hack` can also be configured at nginx level, since i18n route is also configured [there](https://github.com/cofacts/rumors-deploy/blob/master/volumes/nginx/sites-enabled/rumors-site#L42).

What do you think?

MrOrz:
This should be handled in nginx, as discussed in https://g0v.hackmd.io/kJws0_UQSfOAOjOGuXQn1w#PRs

Issue #135, opened 2018-08-25
Landing page not displaying normally on phones


changhc:
The landing page's CSS for mobile devices needs revision.
And maybe a separate style for i18n? See the second screenshot, which was captured with my phone.

![screenshot from 2018-08-26 07-25-28](https://user-images.githubusercontent.com/4591982/44623456-8a96cc00-a901-11e8-9075-0b19b706456a.png)

![screenshot_2018-08-26-07-14-18](https://user-images.githubusercontent.com/4591982/44623445-1825ec00-a901-11e8-81f1-7c273a494a07.png)

MrOrz:
This issue should belong to landing page repo: https://github.com/cofacts/grants-landing-template

PR #191, opened 2019-12-21
Fix CSS in landing page

howawong:
Fixing the CSS issue in mobile phone #135


MrOrz:
LGTM! Thank you so much for the fix!

Issue #190, opened 2019-12-21
QR code on landing page is broken


MrOrz:
Next.js file-loader somehow did not resolve file path for QR code correctly. Needs more investigation.

![image](https://user-images.githubusercontent.com/108608/71303289-a6c15500-23f1-11ea-9a82-1a903a3b6bd9.png)

PR #194, opened 2019-12-21
hi~ g0v Tainan great！

Su-Ya:
the reason is qr-code png size less than url-loader limit setting, so it compile to base64.
use import to compile.

螢幕快照 2019-12-21 下午2 50 05
螢幕快照 2019-12-21 下午2 50 36


MrOrz:
Thank you so much for fixing this!

MrOrz:
This fixes #190

PR #214, opened 2020-02-09
Setup pm2 cluster mode on production

MrOrz:
Currently, site-zh often goes to CPU 100% and becomes extremely slow; in the mean time, our Linode actually has multiple cores. Currently rumors-site only uses 1 core, if we leverage more cores, the CPU usage should hit 100% less often.

This PR:
- Adds a custom server that works exactly the same as `next start`
- Modifies `Dockerfile` so that it runs pm2 and loads the custom server script
- Enables cluster mode so that we can control number of processes using `WEB_CONCURRENCY` env var on docker container

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ WORKDIR /srv/www
 EXPOSE 3000
 
 # Note: using npm start here will cause error when exiting the container
-ENTRYPOINT node_modules/.bin/pm2-runtime ecosystem.config.js
+CMD ["node_modules/.bin/pm2-runtime", "ecosystem.config.js"]
 
 COPY package.json package-lock.json next.config.js ecosystem.config.js server.js ./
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -27,8 +27,10 @@ FROM node:12-stretch-slim
 
 WORKDIR /srv/www
 EXPOSE 3000
-ENTRYPOINT npm start
 
-COPY package.json package-lock.json next.config.js ./
+# Note: using npm start here will cause error when exiting the container
+ENTRYPOINT node_modules/.bin/pm2-runtime ecosystem.config.js
```
LucienLee: how about use cmd?
https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/

MrOrz: Thanks for advice, changed to `CMD`

PR #230, opened 2020-03-04
Improve/article list ui

yanglin5689446:
- Add basic UI to article list items so that when orderBy equals to 'lastRepliedAt', it displays the latest replied message below the original article.

### DIFF
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -1,7 +1,10 @@
 import gql from 'graphql-tag';
 import Link from 'next/link';
 import ArticleInfo from './ArticleInfo';
+import { t } from 'ttag';
 import { listItemStyle } from './ListItem.styles';
+import isValid from 'date-fns/isValid';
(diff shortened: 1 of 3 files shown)


components/ArticleItem.js
```diff
@@ -1,17 +1,25 @@
...
   article,
   read = false, // from localEditorHelperList, it only provide after did mount
   notArticleReplied = false, // same as top
+  showLastReply = true,
   // handleLocalEditorHelperList,
   // isLogin,
 }) {
+  const latestReply = article.articleReplies[0]?.reply;
+  const createdAt = new Date(latestReply.createdAt);
+  const timeAgoStr = formatDistanceToNow(createdAt);
```
MrOrz: (Trivial) When I read this in `ArticleItem`, I would suppose it's `createdAt` of article, instead of latest reply. I suggest using variable names like `latestReplyCreatedAt` and `latestReplyTimeAgoStr` instead

PR #231, opened 2020-03-05
Handle when latestReply is undefined

MrOrz:
Currently visiting [article list](https://cofacts.hacktabl.org/articles) will see 500 internal server error:
image

The root cause is `ArticleItem` trying to read `createAt` on empty reply:
```
cofacts-staging-zh_1_26a0c24950d3 | 2020-03-05 06:04 +00:00: Error while running `getDataFromTree` TypeError: Cannot read property 'createdAt' of undefined
cofacts-staging-zh_1_26a0c24950d3 |     at ArticleItem (/srv/www/.next/server/static/RmJ6dlBKzFB7gLIxTollr/pages/articles.js:331:46)
```

Another bug is, when navigating to article page with existing reply, this error may come up:
```
    'Store error: the application attempted to write an object with no provided ' +
    'id but the store already contains an id of Reply:Hjq2E2UBbZnN2I-EOZtn for ' +
    'this object. The selectionSet that was trying to be written is:\n' +
    '{"kind":"Field","name":{"kind":"Name","value":"reply"},"arguments":[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"text"},"arguments":[],"directives":[]},{"kind":"Field","name":{"kind":"Name","value":"createdAt"},"arguments":[],"directives":[]},{"kind":"Field","name":{"kind":"Name","value":"__typename"}}]}}',
```

This PR fixes the errors above via:
1. early-return when rendering latest reply
2. provide id in reply query

### DIFF
--- a/components/ArticleItem.js
+++ b/components/ArticleItem.js
@@ -8,6 +8,35 @@ import { format, formatDistanceToNow } from 'lib/dateWithLocale';
 // import ArticleItemWidget from './ArticleItemWidget/ArticleItemWidget.js';
 import cx from 'clsx';
 
+function LatestReply({ reply }) {
+  if (!reply) return null;
+
+  const lastRepliedAt = new Date(reply.createdAt);
+  const timeAgoStr = formatDistanceToNow(lastRepliedAt);
+
+  return (
+    <div className="latest-reply">
+      <strong>{t`Latest Reply`}</strong>
+      <br />
+      {reply.text}
+      {isValid(lastRepliedAt) && (
+        <span title={format(lastRepliedAt)}>
+          {' - '}
+          {t`${timeAgoStr} ago`}
+        </span>
+      )}
+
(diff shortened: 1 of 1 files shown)


components/ArticleItem.js
```diff
@@ -56,16 +71,6 @@ export default function ArticleItem({
       </Link>
 
       <style jsx>{listItemStyle}</style>
-      <style jsx>{`
-        .latest-reply {
-          background-color: #64b5f6;
-          padding: 1rem;
-          border-radius: 4px;
-        }
-        .item:hover .latest-reply {
```
MrOrz: @yanglin5689446 Since it's difficult to have cross-component style, and considering that we are revamping UI very soon, I did not migrate this style.

PR #234, opened 2020-03-06
Category list in article page

MrOrz:
Implements Article page UI in #213.

- List article categories under article text & reply request list.
- The person adding the category (`canUpdateStatus === true`) can remove category (has delete button)
- Clicking category label opens a dropdown that has options which
    - links to article list under that category
    - performs upvote and display upvote count
        - If user has upvoted, the option will be disabled.
    - opens downvote dialog, which displays others' reason of downvote and can fill in own downvote reason
        - If user has downvoted, the dialog should include their own submitted reason. They can update their own reason by casting downvote again.
    - "Add category" button is hidden by default. Shown only when env var `PUBLIC_SHOW_ADD_CATEGORY` is set.

## Screenshots

image

### Casting upvote
upvote

### Listing other's reason and casting downvote
Currently the API mock does not contain any reason, thus the list is empty.
downvote

## TODO
The implementation of "Add category" button will be implemented in the next PR.

## Refactor
- export `dataIdFromObject` for mutation updates
- Resolve "Form submission is canceled" warning when submitting feedback reason for article replies.

### DIFF
--- a/components/ArticleCategories.js
+++ b/components/ArticleCategories.js
@@ -1,0 +1,29 @@
+import gql from 'graphql-tag';
+import Chip from '@material-ui/core/Chip';
+
+const ArticleCategoriesData = gql`
+  fragment ArticleCategoriesData on ArticleCategory {
+    articleId
+    categoryId
+    category {
+      title
+      description
+    }
+  }
+`;
+
+function ArticleCategories({ articleId, articleCategories }) {
+  return (
+    <aside>
+      {(articleCategories || []).map(({ categoryId, category }) => (
+        <Chip key={categoryId} label={category.title} />
+      ))}
+    </aside>
+  );
+}
+
+ArticleCategories.fragments = {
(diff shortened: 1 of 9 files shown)


components/ArticleCategories.js
```diff
@@ -40,8 +137,10 @@ function Category({ category }) {
     <>
       <Chip
         className={classes.category}
+        disabled={deletingCategory}
         label={category.title}
         onClick={handleOpen}
+        onDelete={canDelete ? () => deleteCategory() : undefined}
```
yanglin5689446: why not just write `canDelete ? deleteCategory : undefined`

MrOrz: I don't want DOM events being sent to mutation function, thus wrapping with arrow function

MrOrz: Actually I did not test if `Chip`'s `onDelete` passes anything :P

PR #239, opened 2020-03-14
add component "CreateReplyRequestDialog"

Stimim:
please take a look.

### DIFF
--- a/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
+++ b/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
@@ -65,13 +65,20 @@ class CreateReplyRequestDialog extends React.PureComponent {
       text: value,
       disabled: !value || value.length < MIN_REASON_LENGTH,
     });
+
+    // Backup to localStorage
+    requestAnimationFrame(() => (localStorage.text = value));
   };
 
   handleReasonSubmitted = () => {
     this.formRef.current.reset();
+
     this.setState({
+      text: '',
       visible: false,
     });
+
+    requestAnimationFrame(() => (localStorage.text = ''));
   };
 
   showForm = () => {
--- a/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
+++ b/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
@@ -60,13 +60,6 @@ class CreateReplyRequestDialog extends React.PureComponent {
     });
   }
 
-  set(key, value) {
-    this.setState({ [key]: value });
-
-    // Backup to localStorage
-    requestAnimationFrame(() => (localStorage[key] = value));
-  }
-
   handleTextChange = ({ target: { value } }) => {
     this.setState({
       text: value,
@@ -74,7 +67,7 @@ class CreateReplyRequestDialog extends React.PureComponent {
     });
   };
 
-  onReasonSubmitted = () => {
+  handleReasonSubmitted = () => {
     this.formRef.current.reset();
     this.setState({
       visible: false,
@@ -104,6 +97,7 @@ class CreateReplyRequestDialog extends React.PureComponent {
             <textarea
               placeholder="例：我用 OO 關鍵字查詢 Facebook，發現⋯⋯ / 我在 XX 官網上找到不一樣的說法如下⋯⋯"
               onChange={this.handleTextChange}
+              value={text}
             ></textarea>
             <details>
               <summary>送出理由小撇步</summary>
@@ -118,7 +112,7 @@ class CreateReplyRequestDialog extends React.PureComponent {
               articleId={this.props.articleId}
               text={text}
               disabled={disabled}
-              onFinish={this.onReasonSubmitted}
+              onFinish={this.handleReasonSubmitted}
             />
             <button onClick={this.onCancel}>取消</button>
 
--- a/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
+++ b/components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
@@ -16,8 +16,9 @@ const CREATE_REPLY_REQUEST = gql`
     }
   }
 `;
+const MIN_REASON_LENGTH = 80;
 
-function SubmitButton({ articleId, text, disabled }) {
+function SubmitButton({ articleId, text, disabled, onFinish }) {
   const [createReplyRequest] = useMutation(CREATE_REPLY_REQUEST, {
     refetchQueries: ['LoadArticlePage'],
   });
@@ -26,11 +27,12 @@ function SubmitButton({ articleId, text, disabled }) {
     e.preventDefault(); // prevent reload
     if (disabled) return;
     createReplyRequest({ variables: { articleId, reason: text } });
+    onFinish();
   };
 
   return (
     <button onClick={handleSubmit} disabled={disabled}>
-      送出
+      {disabled ? "字數太少，無法送出" : "送出理由"}
     </button>
   );
 }
@@ -43,6 +45,8 @@ class CreateReplyRequestDialog extends React.PureComponent {
     this.state = {
       ...formInitialState,
     };
+
+    this.formRef = React.createRef();
   }
 
   componentDidMount() {
@@ -66,12 +70,23 @@ class CreateReplyRequestDialog extends React.PureComponent {
   handleTextChange = ({ target: { value } }) => {
     this.setState({
       text: value,
-      disabled: !value || value.length < 80,
+      disabled: !value || value.length < MIN_REASON_LENGTH,
     });
   };
 
+  onReasonSubmitted = () => {
+    this.formRef.current.reset();
+    this.setState({
+      visible: false
+    });
+  }
+
   showForm = () => {
     this.setState({ visible: true });
+  };
+
+  onCancel = () => {
+    this.setState({ visible: false });
   };
 
   render = () => {
@@ -80,12 +95,12 @@ class CreateReplyRequestDialog extends React.PureComponent {
     return (
       <div>
         {visible ? null : (
-          <button type="button" onClick={this.showForm}>
-            回覆
+          <button onClick={this.showForm}>
+            增加回報理由
           </button>
         )}
         {!visible ? null : (
-          <form>
+          <form ref={this.formRef}>
             <p>
               請告訴其他編輯：<strong>您為何覺得這是一則謠言</strong>？
             </p>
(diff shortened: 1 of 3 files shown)


components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,125 @@
...
+  mutation CreateReplyRequestFromForm($articleId: String!, $reason: String!) {
+    CreateReplyRequest(articleId: $articleId, reason: $reason) {
+      id
+    }
+  }
+`;
+
+function SubmitButton({ articleId, text, disabled }) {
+  const [createReplyRequest] = useMutation(CREATE_REPLY_REQUEST, {
+    refetchQueries: ['LoadArticlePage'],
```
MrOrz: Please clear the form (get the ref of `<form>` and invoke its [`reset()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reset) ) after the request completes.

Also, please set form state's `visible` to false after submission.

I think putting useMutation directly inside `CreateReplyRequestDialog` and avoid creating an isolate `SubmitButton` component can make things easier, and it is natural for a "Form" component to handle its submission.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  render = () => {
+    const { text, visible, disabled } = this.state;
+
+    return (
+      <div>
+        {visible ? null : (
+          <button type="button" onClick={this.showForm}>
+            回覆
```
MrOrz: Please use "增加回報理由" or something like "我也想知道" as depicted in #13, because asking things here is actually not "回覆".

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,125 @@
...
+    this.setState({ [key]: value });
+
+    // Backup to localStorage
+    requestAnimationFrame(() => (localStorage[key] = value));
+  }
+
+  handleTextChange = ({ target: { value } }) => {
+    this.setState({
+      text: value,
+      disabled: !value || value.length < 80,
```
MrOrz: Lets declare a const for 80 and put it on top part of this file

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const handleSubmit = e => {
+    e.preventDefault(); // prevent reload
+    if (disabled) return;
+    createReplyRequest({ variables: { articleId, reason: text } });
+  };
+
+  return (
+    <button onClick={handleSubmit} disabled={disabled}>
+      送出
```
MrOrz: Lets use "字數太少，無法送出" when `disabled` is `true`, and "送出理由" when `disabled` is `false`.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,125 @@
...
+
+  showForm = () => {
+    this.setState({ visible: true });
+  };
+
+  render = () => {
+    const { text, visible, disabled } = this.state;
+
+    return (
+      <div>
```
MrOrz: I think it looks OK, but currently users cannot hide the "dialog" after opening it.

I suggest either we add a cancel button, or we directly use a [simple dialog](https://material-ui.com/components/dialogs/#simple-dialogs) and provide its `onClose` callback.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
Stimim: I tried to call "useMutation" in different functions of CreateReplyRequestDialog, however, it would complain about "Invalid hook call", and the page won't show up at all.

https://github.com/Stimim/rumors-site/commit/31179aa3e9ecb393c70e757e99b16f7f6c17a6dc

MrOrz: Oh it's because hooks [can only be used in functional components](https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both), so `CreateReplyRequestDialog` must be rewritten to functions in order to use `useMutation`.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,138 @@
...
+
+    // restore from localStorage if applicable.
+    // We don't do this in constructor to avoid server/client render mismatch.
+    //
+    this.setState({
+      text: localStorage.text || text,
+    });
+  }
+
+  set(key, value) {
```
MrOrz: Does this method get invoked?

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,138 @@
...
+      <div>
+        {visible ? null : <button onClick={this.showForm}>增加回報理由</button>}
+        {!visible ? null : (
+          <form ref={this.formRef}>
+            <p>
+              請告訴其他編輯：<strong>您為何覺得這是一則謠言</strong>？
+            </p>
+
+            <textarea
+              placeholder="例：我用 OO 關鍵字查詢 Facebook，發現⋯⋯ / 我在 XX 官網上找到不一樣的說法如下⋯⋯"
```
MrOrz: The text in `localStorage` is not provided as default props. Although it's loaded from `localStorage` to state, the user won't know that previous text is loaded already.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
MrOrz: I am ok with current settings though, since it's working now

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  handleTextChange = ({ target: { value } }) => {
+    this.setState({
+      text: value,
+      disabled: !value || value.length < MIN_REASON_LENGTH,
+    });
+  };
+
+  onReasonSubmitted = () => {
```
MrOrz: Let's use `handleReasonSubmit` because it is a handler method :D

We use `onXXX` for event handling props and use `handleXXX` for methods that actually implements event handling.

This helps a lot when we read code in render function, especially when we need to pass handlers from parent to children in complex controls -- if we put `handleXXX` then it means that we do some processing in the component we currently read, and if we see `<ChildComp onXXX={onXXX} />` it must be passing parent handlers `this.props.onXXX` to children.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -60,25 +60,25 @@ class CreateReplyRequestDialog extends React.PureComponent {
...
       disabled: !value || value.length < MIN_REASON_LENGTH,
     });
+
+    // Backup to localStorage
+    requestAnimationFrame(() => (localStorage.text = value));
   };
 
-  onReasonSubmitted = () => {
+  handleReasonSubmitted = () => {
     this.formRef.current.reset();
```
MrOrz: Previously I suggested adding `reset()` because `<textarea>` was a [uncontrolled input](https://reactjs.org/docs/uncontrolled-components.html). However, in recent changes you provide `value` to `<textarea>`, which is turned into a [controlled input](https://reactjs.org/docs/forms.html#controlled-components).

I think we should take one of the following:

- If we want to make the `<textarea>` a controlled input, setting `this.state.text` to `''` can do the clearing. `reset()` and the `formRef` are no longer needed and should be removed.
- If we want to revert `<textarea>` back to an uncontrolled input, then change `value={text}` below to `defaultValue={text}`.

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
```diff
@@ -0,0 +1,139 @@
...
+    this.setState({ visible: false });
+  };
+
+  render = () => {
+    const { text, visible, disabled } = this.state;
+
+    return (
+      <div>
+        {visible ? null : <button onClick={this.showForm}>增加回報理由</button>}
+        {!visible ? null : (
```
MrOrz: Why not combine the two expressions to `{visible ? <form>...</form> : <button>...</button> }` ?

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
Stimim: I think this first depends on UX.
We make textarea controlled input because we want to save and reload the previous text.
So the question is whether we want to cache user's input or not.
This feature is implemented in ReplyForm, do users like that feature?

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
Stimim: I split them into two statements because I think it would be hard to read when they are combined (see below).
Another idea is moving them into `renderForm` and `renderShowButton` respectively, so the `?:` expression would be shorter.

```
    return (
      <div>
        {visible ? (
          <form ref={this.formRef}>
            <p>
              請告訴其他編輯：<strong>您為何覺得這是一則謠言</strong>？
            </p>

            <textarea
              placeholder="例：我用 OO 關鍵字查詢 Facebook，發現⋯⋯ / 我在 XX 官網上找到不一樣的說法如下⋯⋯"
              onChange={this.handleTextChange}
              value={text}
            ></textarea>
            <details>
              <summary>送出理由小撇步</summary>
              <ul>
                <li>闡述更多想法</li>
                <li>去 google 查查看</li>
                <li>把全文複製貼上到 Facebook 搜尋框看看</li>
                <li>把你的結果傳給其他編輯參考吧！</li>
              </ul>
            </details>
            <SubmitButton
              articleId={this.props.articleId}
              text={text}
              disabled={disabled}
              onFinish={this.handleReasonSubmitted}
            />
            <button onClick={this.onCancel}>取消</button>

            <style jsx>{`
              textarea {
                width: 100%;
                height: 5em;
              }
            `}</style>
          </form>
        ) : (
          <button onClick={this.showForm}>增加回報理由</button>
        )}
      </div>
    );
```

components/CreateReplyRequestDialog/CreateReplyRequestDialog.js
MrOrz: Both controlled and uncontrolled input can implement automayic localStorage cache. The key difference between controlled and uncontrolled is `value` prop, not `onChange` prop.

PR #241, opened 2020-03-22
Feature/new navbar ui

yanglin5689446:
[card](https://github.com/orgs/cofacts/projects/5#card-34956857)
navbar and layout UI redesign


components/AppLayout/AppHeader.js
```diff
@@ -1,64 +1,281 @@
...
+            </>
+          ) : (
+            <Button
+              onClick={openLoginModal}
+              size="small"
+              variant="medium"
+            >{t`Login`}</Button>
+          ))}
+      </div>
+      {!isDesktop && (
```
MrOrz: Let's use `<Box display={...}>...</Box>` or [`<Hidden implementation="css"></Hidden>`](https://material-ui.com/api/hidden/) to replace `useMediaQuery` here, as suggested by [useMediaQuery doc](https://material-ui.com/components/use-media-query/#server-side-rendering).

components/AppLayout/AppHeader.js
```diff
@@ -1,64 +1,281 @@
...
-        }
+const MENU_BUTTON_WIDTH = 48;
+
+const useStyles = makeStyles({
+  root: {
+    position: 'sticky',
+    height: NAVBAR_HEIGHT + TABS_HEIGHT,
+    top: 0,
+    zIndex: 10,
+    '@media(min-width: 992px)': {
```
MrOrz: We can [customize breakpoint values](https://material-ui.com/customization/default-theme/?expand-path=$.breakpoints.values) (maybe change `md` to `992`) so that we can use `theme.breakpoints.up('md')` here instead.

@LucienLee what do you think about changing [Material UI's default breakpoints](https://material-ui.com/customization/breakpoints/#breakpoints)? Should we bring `md` up to `992` or bring `lg` down to `992`?

yanglin5689446: 992 is defined by bootstrap, which is different from material-ui, using breakpoints defined by material-ui should be preferred, I'll fix this.

components/AppLayout/AppFooter.js
```diff
@@ -1,36 +1,126 @@
...
+} from 'constants/urls';
+import NavLink from 'components/NavLink';
+import { withStyles, makeStyles } from '@material-ui/core/styles';
+import MailIcon from '@material-ui/icons/Mail';
+
+const useStyles = makeStyles({
+  first: {
+    display: 'flex',
+    justifyContent: 'center',
+    // @todo: use material-ui builtin palette color
```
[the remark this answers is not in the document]
yanglin5689446: already fixed in 0d2085e

components/AppLayout/AppHeader.js
MrOrz: Yeah, but I remember @LucienLee is referring to bootstrap breakpoints in his design. Not sure which breakpoint setup we should follow.

components/NavLink.js
```diff
@@ -0,0 +1,34 @@
...
+  target,
+  rel,
+  children,
+}) {
+  const router = useRouter();
+  return external ? (
+    <a
+      href={href}
+      target={target || '_blank'}
+      rel={rel || 'noopener noreferrer'}
```
MrOrz: I think target-blank and no-referrer by default is a bit opinionated. Since `NavLink` is mostly used for navigation within Cofacts, we should not set default for these values.

Although we are removing the defaults so it's not used here, I would like to mention that it is preferred to set default props (if any) using object destructuring default value `function NavLink({prop1 = 'defaultValue1'})`. It has the following benefit:
1. Default value declared with prop, more declarative
2. Editors that pick up function definition (like VS Code) can show default when invoking the function / component
3. [Storybook addon info](https://github.com/storybookjs/storybook/tree/master/addons/info) can automatically pick up default values in its props table

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+import AccountCircleOutlinedIcon from '@material-ui/icons/AccountCircleOutlined';
+import ExitToAppRoundedIcon from '@material-ui/icons/ExitToAppRounded';
+import InfoIcon from '@material-ui/icons/Info';
+import { NAVBAR_HEIGHT, TABS_HEIGHT } from 'constants/size';
 import { EDITOR_FACEBOOK_GROUP, PROJECT_HACKFOLDR } from 'constants/urls';
-import Link from 'next/link';
-import UserName from './UserName';
+import * as Widgets from './Widgets';
+import desktopLogo from './images/logo-desktop.png';
+import mobileLogo from './images/logo-mobile.png';
```
MrOrz: I wonder if SVG files are smaller than png files?

components/NavLink.js
MrOrz: Sorry I missed the `external` part. Please ignore the comment above

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+            <Button
+              onClick={openLoginModal}
+              size="small"
+              className={classes.login}
+            >{t`Login`}</Button>
+          )}
+        </Box>
+      </div>
+      <Box display={['block', 'block', 'none']}>
+        <div className={classes.flex}>
```
MrOrz: Can use `<Box display={['flex', 'flex', 'none']}></Box>` instead of 2 nested `div`s, simplify the DOM

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+          <a href="/">
+            <img
+              className={classes.logo}
+              src={mobileLogo}
+              srcSet={`${desktopLogo} ${theme.breakpoints.values.md}w`}
+              alt=""
+            />
+          </a>
+          <Box display={['none', 'none', 'inline']}>
```
MrOrz: If we use fragment as `<Links>`'s root component, we can use `<Box display={['none', 'none', 'flex']}>` here to make a flex container for all items

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+      height: 'auto',
+      width: 'auto',
+      fontSize: 20,
+      padding: '0 10px',
+    },
+  },
+  tab: {
+    padding: 8,
+    flex: '1',
+    lineHeight: `calc(${TABS_HEIGHT}px - 8px * 2)`,
```
MrOrz: If we can ensure all `.tab`s are flex child (with `align-items` being `stretch`), I prefer using flexbox to vertically center the text, i.e. `display: flex` + `align-items: center` here, rather than using `lineHeight` + `calc`. In this way we don't need to keep paddings & height in sync.

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+  return (
+    <header className={classes.root}>
+      <div className={classes.top}>
+        <div className={classes.flex}>
+          <a href="/">
+            <img
+              className={classes.logo}
+              src={mobileLogo}
+              srcSet={`${desktopLogo} ${theme.breakpoints.values.md}w`}
+              alt=""
```
MrOrz: It's showing desktop logo on mobile screen:
image

I think its because the lack of `sizes` property documented here: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Different_sizes

Another thing is, the logo looks blurry. I think using SVGs instead of PNGs should fix that.

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
-      `}</style>
+function AppHeader({ onMenuButtonClick, user, openLoginModal, logout }) {
+  const [anchor, setAnchor] = useState(false);
+  const classes = useStyles();
+  const theme = useTheme();
+
+  const openProfileMenu = useCallback(e => setAnchor(e.currentTarget), [
+    anchor,
+  ]);
+  const closeProfileMenu = useCallback(() => setAnchor(null), [anchor]);
```
MrOrz: I think `anchor` is not needed in dependency list of `openProfileMenu` and `closeProfileMenu`.

Actually I think it's OK to not use `useCallback` here at all, since entire `<AppHeader>` is pure-rendered and thus not updated often.

components/AppLayout/AppHeader.js
```diff
@@ -1,66 +1,244 @@
...
+            <Links classes={classes} />
+          </Box>
+        </div>
+        {/* GlobalSearch not fully implemented yet */}
+        {/* <GlobalSearch />*/}
+        <Box display={['none', 'none', 'block']}>
+          {user?.name ? (
+            <>
+              <Widgets.Avatar user={user} size={40} onClick={openProfileMenu} />
+              <Menu
```
MrOrz: We may need to hunt down the root cause of this warning:
image

components/AppLayout/AppHeader.js
yanglin5689446: seems like png is smaller
截圖 2020-03-31 下午8 05 52

components/AppLayout/AppHeader.js
yanglin5689446: not sure why it looks so blur in original size, but it looks decent when zoomed in, really weired
截圖 2020-03-31 下午8 35 57

yanglin5689446: also the bug of misplacing image was fixed, but I use the picture method mentioned here instead: 
https://cythilya.github.io/2018/08/24/responsive-images/

components/AppLayout/AppLayout.js
```diff
@@ -42,11 +84,25 @@ function AppLayout({ children }) {
...
   return (
     <Fragment>
-      <AppHeader />
-      {isRouteChanging && <LinearProgress classes={classes} />}
-      <Container>{children}</Container>
+      <AppHeader
+        onMenuButtonClick={toggleSidebar}
+        user={data?.GetUser}
+        openLoginModal={openLoginModal}
+        logout={logout}
```
MrOrz, line 91: I think using `onLogout={logout}` / `onLoginModalOpen={openLoginModal}` adheres to the convention more. The same goes to `AppSidebar`.

components/AppLayout/Widgets/Level.js
```diff
@@ -0,0 +1,50 @@
...
+      content: '""',
+    },
+  },
+}));
+
+// @todo: material-ui menu component implicitly uses ref,
+// not passing ref to component causes runtime warning.
+// maybe we should find a better solution to deal with this.
+// eslint-disable-next-line react/display-name
+export default React.forwardRef(function Level(
```
MrOrz, line 40: Let's fix the display name:

```javascript
const Level = React.forwardRef(function(...) {...});
Level.displayName = Level;
export default Level; 
```

Then we can get rid of the `eslint-disable-next-line`

components/AppLayout/Widgets/LevelProgressBar.js
```diff
@@ -0,0 +1,43 @@
...
+    height: 14,
+  },
+  progress: {
+    display: 'block',
+    margin: 1,
+    width: percent => `${percent}%`,
+    backgroundColor: theme.palette.primary[500],
+    borderRadius: 10,
+    height: 10,
+    content: ' ',
```
MrOrz, line 19: Not pseudo element, extra `content` property

components/AppLayout/Widgets/index.js
```diff
@@ -0,0 +1,7 @@
+import Avatar from './Avatar';
+import Level from './Level';
+import LevelProgressBar from './LevelProgressBar';
+
+export { Avatar };
```
MrOrz, line 5: Can use `export { Avatar, Level, LevelProgressBar }` instead.

components/NavLink.js
```diff
@@ -0,0 +1,29 @@
...
+import { useRouter } from 'next/router';
+import cx from 'clsx';
+
+function NavLink({
+  href,
+  className,
+  activeClassName,
+  external,
+  target = '_blank',
+  rel = 'noopener noreferrer',
```
MrOrz, line 12: Sorry, my last comment on this component were totally wrong.

These defaults only makes sense only when `external` is `true`, they should be to put `_blank` and `noopener noreferrer` inside the function body, otherwise it will apply to `NavLink`s even when `external` is `false`.

yanglin5689446: In fact when `external` is set to false the target & rel won't be passed down to DOM, so either way would work I guess. will adapt initial method anyway.

MrOrz:
@yanglin5689446 有 conflict ㄛ

PR #247, opened 2020-04-13
Feature/new pages ui

yanglin5689446:
- [x] new article list UI
- [x] new replies page UI
- [x] search bar
- [x] hoax for you page
- [x] search result page(wip)
- [ ] paginator


MrOrz:
@yanglin5689446 currently the file diff contains the changes from merged new-navbar-ui branch, which is pretty difficult to review. Would you rebase the commits?
image

components/ArticleItem.js
```diff
@@ -1,76 +1,154 @@
...
+      <ArticleInfo article={article} />
+      <div className={classes.flex}>
+        {showReplyCount && (
+          <div className={classes.infoBox}>
+            <div>
+              <h2>{~~replyCount}</h2>
+              <span>{t`replies`}</span>
+            </div>
+            <div>
+              <h2>{~~replyRequestCount}</h2>
```
MrOrz, line 119: I prefer `Math.floor` over `~~`.

Actually for this scenario, both `replyCount` and `replyRequestCount` cannot be float. If we are converting something like `null` to 0, we should use [unary plus (`+`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Unary_plus) instead.

components/Filters.js
```diff
@@ -0,0 +1,193 @@
...
+    Object.fromEntries(options.map(option => [option.value, !!option.selected]))
+  );
+  const [expand, setExpand] = useState(false);
+
+  const onOptionClicked = useCallback(
+    value => () => {
+      if (multiple) {
+        const newState = { ...selected, [value]: !selected[value] };
+        setSelected(newState);
+        onChange(newState);
```
MrOrz, line 120: Personally I think it is more natural if we pass a list of selected values as `onChange()` argument, instead of a map with true/false, for a multi-select `<Filter>`.

Passing the state outside `Filter` is like exposing the inner working (structure of internal state) of the component, whenever a developer uses `<Filter onChange>` they will need to know and deal with this particular structure.

components/ReplyItem.js
```diff
@@ -1,55 +1,426 @@
...
+      <Box pr="14px" display={['block', 'block', 'none']}>
+        <Avatar user={user} size={30} />
+        {/*
+          <div title={TYPE_NAME[replyType]}>{TYPE_ICON[replyType]}</div>
+        */}
+      </Box>
+      <Box py="12px" flexGrow={1}>
+        <div
+          className={classes.replyType}
+        >{t`${user.name} consider this ${TYPE_NAME[replyType]}`}</div>
```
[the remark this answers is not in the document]
MrOrz: Another thing is, I am wondering if we should use `reply`'s user name, or `articleReply`'s user name, since they can be different users, and the original author of the reply could think differently from the person that connects their reply to some other messages.

Discussion opened here: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO?node-id=79:9#23909174

components/Filters.js
```diff
@@ -0,0 +1,193 @@
...
+              />
+            ))
+        )}
+      </div>
+      {expandable && (
+        <div className={classes.expand}>
+          <button type="button" onClick={() => setExpand(e => !e)}>
+            {expand ? (
+              <>
+                {t`Expand Less`}
```
MrOrz, line 163: Let's use `Collapse`

components/Filters.js
```diff
@@ -0,0 +1,193 @@
...
+                <ExpandLessIcon />
+              </>
+            ) : (
+              <>
+                {t`Expand More`}
```
MrOrz, line 168: Let's use `Expand`

lib/theme.js
```diff
@@ -36,7 +36,7 @@ const theme = createMuiTheme({
       main: red.A400,
     },
     background: {
-      default: '#fff',
+      default: '#e5e5e5',
```
MrOrz, line 39: I see why you are using `#e5e5e5`, it's probably a bug in Figma. I think`#f5f5f5` is the correct one.

bg

pages/hoax-for-you.js
```diff
@@ -0,0 +1,44 @@
...
+  const queryString = querystring.stringify(query);
+
+  return (
+    <AppLayout>
+      <Head>
+        <title>{t`Hoax for you`}</title>
+        <link
+          rel="alternate"
+          type="application/rss+xml"
+          href={`${PUBLIC_URL}/api/articles/rss2?${queryString}`}
```
MrOrz, line 24: I think we should remember to revisit RSS feed URLs later (in future PRs), because currently these `queryString` does not reflect the actual query.

pages/hoax-for-you.js
```diff
@@ -0,0 +1,44 @@
...
+          href={`${PUBLIC_URL}/api/articles/atom1?${queryString}`}
+        />
+      </Head>
+      <ArticlePageLayout
+        title={t`Hoax for you`}
+        filters={{
+          status: false,
+          consider: false,
+          category: true,
+        }}
```
MrOrz, line 38: Seems that the API needed for this page is not ready yet. Let's revisit it in future PRs.

pages/search.js
```diff
@@ -0,0 +1,99 @@
...
+        />
+      </Head>
+      <div className={classes.jumbotron}>
+        <Container>
+          <Box py="56px">
+            <h2 className={classes.search}>{t`Search`}</h2>
+          </Box>
+        </Container>
+        <Container>
+          <Box display="flex">
```
MrOrz, line 74: I think we should still use [material-ui tabs](https://material-ui.com/components/tabs/) for more consistent look across tabs in the UI.

components/AppLayout/AppHeader.js
```diff
@@ -104,9 +106,33 @@ const useStyles = makeStyles(theme => ({
     borderRadius: 70,
     border: `1px solid ${theme.palette.secondary[500]}`,
   },
+  withLabel: {
+    position: 'relative',
+  },
+  label: {
+    [theme.breakpoints.up('md')]: {
```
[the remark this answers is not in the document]
yanglin5689446: it seems like there is no label in original design for mobile. btw I think I should rename this to badge according to material ui logic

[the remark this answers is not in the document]
MrOrz: I think we can use the same red badge in mobile as well. LINE on Android also have red badges in its bottom nav 📱

components/ReplyItem.js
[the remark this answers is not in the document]
yanglin5689446: about reply's user name I'll take a note and remain the same for the time being, because I haven't figured out the "reference another reply" logic yet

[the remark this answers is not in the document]
MrOrz: Got it, let's save this to future PRs

components/ReplyItem.js
```diff
@@ -1,55 +1,426 @@
...
+      >
+        <button
+          type="button"
+          className={classes.closeButton}
+          onClick={closeReasonsPopover}
+        >
+          <CloseIcon />
+        </button>
+        {text}
+        <div className={classes.tabs}>
```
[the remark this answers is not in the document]
yanglin5689446: because that for the current mockup, there will be custom svg icons in tab, and I found it might be difficult adapting tabs of material-ui in this case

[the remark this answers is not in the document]
MrOrz: Hmm but Material-UI tabs already support icons + text, it should be doable by customizing Tab's `wrapper` to make it align horizontally, or use `position: absolute` on icons and use right-padding on `wrapper` to make space.

image

pages/search.js
yanglin5689446: because material-ui tabs use `onChange` with `tabIndex` as argument, but in this case these 2 tabs are for navigating to different page, and the logic isn't quite relate to the tabIndex.

of course the we can do something like:
```
handleChange(index){
  switch(index) {
    case 0: return navigate('messages');
    case 1: return navigate('replies');
  }
}
```
but personally I think it's a bit more confusing then current method

MrOrz: Actually we can provide `value` prop to `Tab`. In this way, we can do this:
```
<Tabs onChange={(e, page) => navigate(page)}>
  <Tab value="messages" />
  <Tab value="replies" />
</Tabs>
```

yanglin5689446: wow didn't notice there is an API for this, will fix later

components/ReplyFeedback.js
```diff
@@ -34,24 +181,24 @@ const ArticleReplyFeedbackForUser = gql`
 const ArticleReplyFeedbackData = gql`
   fragment ArticleReplyFeedbackData on ArticleReply {
     ...ArticleReplyFeedbackForUser
-    user {
-      id
-    }
     positiveFeedbackCount
     negativeFeedbackCount
+    ownVote
     feedbacks {
```
MrOrz, line 187: You also need `vote` under each feedbacks, because the you are using `vote` to move feedbacks in the right tab.

Currently both tabs are empty because no feedback has `vote` field......
image

components/ReplyFeedback.js
```diff
@@ -70,151 +217,196 @@ const VOTE_REPLY = gql`
...
-      {!isOwnArticleReply && <label>{t`Is the reply helpful?`}</label>}
-      <span className="vote-num">{positiveFeedbackCount}</span>
+    <div className={cx(classes.root, className)}>
       <button
-        className="btn-vote"
-        onClick={() => {
-          voteReply({ variables: { articleId, replyId, vote: 'UPVOTE' } });
-        }}
-        disabled={isOwnArticleReply || isVotingReply}
+        className={cx(classes.vote, ownVote === 'UPVOTE' && classes.voted)}
```
MrOrz, line 279: Inconsistent size of thumb-up/down button and the reason button in mobile view.

image

Corresponding mockup:
image

pages/search.js
```diff
@@ -0,0 +1,92 @@
...
+            textColor="primary"
+            aria-label="tabs"
+          >
+            <CustomTab value="messages" label={t`Messages`} />
+            <CustomTab value="replies" label={t`Replies`} />
+          </Tabs>
+        </Container>
+      </div>
+      <div className={classes.content}>
+        <ArticlePageLayout q={{ [query.type]: query.q }} />
```
MrOrz, line 86: Search result for replies is different from mockup. It should be listing replies that has the query text, rather than articles that has the query text.

Mockup:
image

Current:
image


This may be huge so can include in future PRs.

components/AppLayout/AppHeader.js
```diff
@@ -106,7 +108,19 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
-const Links = ({ classes }) => (
+const LIST_UNSOLVED_ARTICLES = gql`
+  query ListArticles {
+    ListArticles(filter: { replyCount: { EQ: 0 } }) {
+      totalCount
+    }
```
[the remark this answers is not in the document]
yanglin5689446: https://github.com/orgs/cofacts/projects/5#card-37456782
added here

pages/search.js
yanglin5689446: card added here: https://github.com/orgs/cofacts/projects/5#card-34990256

components/ReplyFeedback.js
```diff
@@ -40,6 +40,14 @@ const useStyles = makeStyles(theme => ({
   voted: {
     border: `1px solid ${theme.palette.primary[500]}`,
     color: theme.palette.primary[500],
+    '& $outlinedIcon': {
+      color: 'transparent',
+      stroke: theme.palette.primary[500],
```
MrOrz, line 45: 其實我是指左邊兩顆 vote 按鈕，與右邊灰底的顯示部分不一樣高
image

Mockup 裡，雖然 mobile / desktop 下的 button 高度不同，但左右兩塊都是等高的。

components/ReplyItem.js
```diff
@@ -76,14 +84,8 @@ function ReplyItem({
...
     <div className={classes.root}>
-      <Box p="24px" display={['none', 'none', 'block']}>
-        <Avatar user={user} size={72} />
-        {/*
-          <div title={TYPE_NAME[replyType]}>{TYPE_ICON[replyType]}</div>
-        */}
-      </Box>
-      <Box pr="14px" display={['block', 'block', 'none']}>
-        <Avatar user={user} size={30} />
+      <Box p={{ xs: '12px 14px 0 0', md: '24px' }}>
```
MrOrz, line 87: mobile 下， avatar 應該要跟右側文字置中。目前這樣好像有點太低了
image

pages/replies.js
```diff
@@ -1,286 +1,45 @@
...
+          href={`${PUBLIC_URL}/api/articles/atom1?${queryString}`}
+        />
+      </Head>
+      <ArticlePageLayout
+        title={t`Replies`}
+        articleDisplayConfig={{
+          isLink: false,
+          showLastReply: true,
+          showReplyCount: false,
+        }}
```
MrOrz, line 39: Actually, replies page has different filter from article page.

Mockup:
image

Spec: https://g0v.hackmd.io/ZZWHWi2BTuyhkSWzAvKLAw#%E9%81%8E%E6%BF%BE%E9%81%B8%E9%A0%85%EF%BC%9A

The API for "我查核過" and reply type filtering is still on the way, but "還未有有效查核", "熱門回報" and "熱門討論" are legit `ListArticle` filters already (`hasPositiveFeedbackArticleReply`, `replyRequestCount` and `replyCount` respectively).

Considering the size of this change, I think we can handle this in future PRs, may need a ticket for this.

yanglin5689446: card added here: https://github.com/orgs/cofacts/projects/5#card-37494391

PR #248, opened 2020-04-30
Fix/text expansion

yanglin5689446:
merge TextExpansion & old ExpandableText components
should merge [247](https://github.com/cofacts/rumors-site/pull/247) first


MrOrz:
Merge base temporarily changed to new page UI for correct diff. Will change back to dev benefits merge.

components/ExpandableText.js
```diff
@@ -1,62 +1,107 @@
...
+  root: {
+    border: 0,
+    color: '#2079F0',
+    outline: 'none',
+    cursor: 'pointer',
+    background: 'transparent',
+  },
+})(({ classes, toggleExpand, expanded }) => (
+  <button className={classes.root} onClick={toggleExpand}>
+    ({expanded ? t`Show Less ▲` : t`Show More ▼`})
```
MrOrz, line 31: We can use this so that ▲ ▼ translators don't need to handle the triangles:
```javascript
({expanded ? t`Show Less` + ' ▲' : t`Show More` + ' ▼'})
```

or

```javascript
({expanded ? `${t`Show Less`} ▲` : `${t`Show More`}  ▼`})
```

components/ExpandableText.js
```diff
@@ -1,62 +1,107 @@
...
-          {this.renderToggleButton()}
-        </div>
-      );
+  useEffect(() => {
+    if (lineClamp) {
+      cloneAndComputeHeight();
     }
+  }, []);
 
+  const renderWithLineChampLimits = () => {
```
MrOrz, line 61: Not sure why, but there are some weird cases that need to trace code to figure out why

image

The article content is `555`. Can be found by searching `555` in page.

Other similar scenario are:
image
image

MrOrz:
I suggest adding the component to storybook so that we can get visual documentation & snapshot testing

components/ExpandableText.js
yanglin5689446: this seems like the old behavior of original commit, but I rebase the commit, it should look like these now: 
截圖 2020-05-06 上午11 10 29
截圖 2020-05-06 上午11 10 14

MrOrz: It seems that it only happens on specific articles that has only 1 line, and the line is super short, on desktop.

You can search for `555` on [staging](https://cofacts.hacktabl.org/replies) to see the following sample

image

MrOrz: Confirmed that its' fixed after rebase.

PR #249, opened 2020-05-06
Feature/new article details page ui

yanglin5689446:
New article details page UI [mockup](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=681%3A0)


MrOrz:
Please `npm run lint:fix` first. https://travis-ci.org/github/cofacts/rumors-site/builds/684561774

components/ArticleReply/CopyButton.js
```diff
@@ -0,0 +1,22 @@
...
+const CopyButton = React.memo(({ content = '', onClick = () => {} }) => {
+  const copyBtnRef = useRef(null);
+
+  useEffect(() => {
+    const clipboard = new ClipboardJS(copyBtnRef.current, {
+      text: () => content,
+    });
+    clipboard.on('success', () => onClick());
+    return () => clipboard.destroy();
+  }, []);
```
MrOrz, line 15: Suggest adding `copyBtnRef.current` and `content` to dependency array

MrOrz:
Deployed to https://cofacts.hacktabl.org/articles so that everyone can inspect.
[Docker image tags](https://hub.docker.com/r/cofacts/rumors-site/tags): `pr249-en` and `pr249-tw`

components/ReplyForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+    <Box py={2}>
+      <Box display="flex" justifyContent="space-between">
+        <label className={classes.label} htmlFor="reference">
+          <strong>
+            {replyType === 'OPINIONATED'
+              ? '請提供與原文「不同觀點」的文章連結，促使讀者接觸不同意見'
+              : '資料來源'}
+          </strong>
+        </label>
+        <Hint>{t`Multiple sources will be separated by line break for better line user experience`}</Hint>
```
MrOrz, line 110: Copy 請改成
> Inserting blank lines between reference items can improve readability in LINE.

cc/ @LucienLee

components/ArticleReply/index.js
```diff
@@ -0,0 +1,242 @@
...
+            pollingId={replyId}
+          />
+        </section>
+      );
+    }, [replyType, reply]);
+
+    const authorElem = renderAuthor();
+
+    return (
+      <li className={classes.root}>
```
MrOrz, line 207: 1. Seems that this line does not appear? Currently there is no separation between replies, which is a bit hard to read. image
2. According to [mockup](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=681%3A0) the horizontal line between replies should be solid lines.

components/ArticleReply/index.js
```diff
@@ -0,0 +1,242 @@
...
+        case 'RUMOR':
+          return '#FB5959';
+        default:
+          return theme.palette.common.black;
+      }
+    },
+  },
+  content: {
+    padding: '17px 0',
+    borderBottom: `1px dotted ${theme.palette.secondary[100]}`,
```
MrOrz, line 41: The line on the top of Reference should be `dashed` not `dotted`

components/ArticleReply/index.js
```diff
@@ -0,0 +1,242 @@
...
+import ReplyFeedback from 'components/ReplyFeedback';
+import EditorName from 'components/EditorName';
+import Hyperlinks from 'components/Hyperlinks';
+import Avatar from 'components/AppLayout/Widgets/Avatar';
+import ReplyInfo from 'components/ReplyInfo';
+import ReplyActions from './ReplyActions';
+import ReplyShare from './ReplyShare';
+
+const useStyles = makeStyles(theme => ({
+  root: {
```
MrOrz, line 20: This class does not seem to have any effect

components/CreateReplyRequestForm/CreateReplyRequestForm.js
```diff
@@ -0,0 +1,238 @@
...
+          <div className={classes.buttonGroup}>
+            <button
+              type="button"
+              className={cx(showForm && 'active')}
+              onClick={() => setShowForm(!showForm)}
+            >{t`I second that`}</button>
+            <button
+              type="button"
+              className={cx(requestedForReply && 'active')}
+            >{t`I also wanna know`}</button>
```
MrOrz, line 198: "I second that" button toggles the form that submits `replyRequest`, thus I think the state `requestedForReply` should link to that button. ( @LucienLee , do we have different state for "I second that" button if the user has requested reply? )

I don't see "I also wanna know" button in mockup now

components/CreateReplyRequestForm/CreateReplyRequestForm.js
```diff
@@ -0,0 +1,238 @@
...
+    const [showForm, setShowForm] = useState(false);
+    const [text, setText] = useState('');
+
+    const [shareAnchor, setShareAnchor] = useState(null);
+
+    useEffect(() => {
+      // restore from localStorage if applicable.
+      // We don't do this in constructor to avoid server/client render mismatch.
+      //
+      setText(localStorage.text || text);
```
MrOrz, line 130: Actually this collides with reply's localstorage cache. This is a bug that can be perceived in current production as well. Changing `text` to some other key should fix the bug; we can also open a separate issue, since this is not in the scope of this PR actually.

螢幕快照 2020-05-16 下午6 53 59

components/Hyperlinks.js
```diff
@@ -1,7 +1,69 @@
...
+    paddingLeft: 10,
+    paddingRight: 15,
+    minHeight: 40,
+    justifyContent: 'space-between',
+    '& .summary': {
+      color: theme.palette.secondary[500],
+      maxHeight: 40,
+      overflow: 'hidden',
+      margin: 0,
+      display: 'flex',
```
MrOrz, line 48: This `flex` is not needed.
Instead, I think we can add `-webkit-line-clamp` (and other properties to make it work) https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp so that it displays `...` when text is truncated.

components/Hyperlinks.js
```diff
@@ -1,7 +1,69 @@
...
+      marginRight: 10,
```
MrOrz, line 49: When there is no image on the right, we don't need `marginRight`. Therefore I suggest we should set `margin-left` on image instead.

components/Hyperlinks.js
```diff
@@ -1,7 +1,69 @@
...
+    margin: '0 8px 8px 0',
+    background: theme.palette.secondary[50],
+    borderRadius: 8,
+    '& h1': {
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      margin: 0,
+      marginBottom: 10,
+      maxWidth: 'inherit',
```
MrOrz, line 18: font-size should be 14. Currently it's slightly larger than 14px due to Google chrome's user agent stylesheet
image

components/RelatedReplies.js
```diff
@@ -1,12 +1,57 @@
...
+    marginLeft: 0,
+  },
+  title: {
+    marginBottom: '.5rem',
+  },
+  submit: {
+    borderBottomLeftRadius: 7,
+    borderBottomRightRadius: 7,
+    backgroundColor: theme.palette.primary[500],
+    color: theme.palette.common.white,
```
MrOrz, line 47: I think that instead of customizing a button from browser style into primary button, we should use [Material UI Contained Buttons with `disableElevation`](https://material-ui.com/components/buttons/#contained-buttons). It has better handle over hover & active states as well.

components/ArticleReply/index.js
```diff
@@ -0,0 +1,242 @@
...
+
+      return (
+        <Box component="footer" display="flex" py={2}>
+          {showFeedback && (
+            <ReplyFeedback
+              articleId={articleId}
+              replyId={replyId}
+              positiveFeedbackCount={positiveFeedbackCount}
+              negativeFeedbackCount={negativeFeedbackCount}
+              feedbacks={feedbacks}
```
MrOrz, line 137: I found a weird case that my feedback is visible in API but not in UI.

螢幕快照 2020-05-17 上午12 48 24

pages/article/[id].js
```diff
@@ -146,97 +237,124 @@ function ArticlePage() {
...
+          </Box>
+        </div>
+
+        <div className={cx(classes.card, classes.aside)}>
+          <h4>{t`Similar messages`}</h4>
+          {similarArticles.length ? (
+            <Box
+              display="flex"
+              flexDirection={{ xs: 'row', md: 'column' }}
+              overflow="scroll"
```
Comment (MrOrz) [inline pages/article/[id].js:337]: Why overflow scroll here? It's introducing always-on horizontal & vertical scrolls.

What can we get from `overflow: scroll`?

image

components/NewReplySection/index.js
```diff
@@ -0,0 +1,160 @@
...
+      relatedArticleReplies,
+      handleSubmit,
+      handleConnect,
+      connectingReply,
+      creatingReply,
+      existingReplyIds,
+    };
+
+    return (
+      <form onSubmit={handleSubmit}>
```
MrOrz, line 137: Nested `form` is introduced here, it's not valid DOM structure. Please remove one of the nesting form.

image

components/NewReplySection/ReplyForm/ReasonEditor.js
```diff
@@ -0,0 +1,156 @@
...
+    '& > *:not(first-child)': {
+      borderLeft: `1px solid ${theme.palette.secondary[100]}`,
+    },
+  },
+  suggestions: {
+    background: theme.palette.common.white,
+    color: theme.palette.secondary[300],
+    padding: 12,
+    whiteSpace: 'nowrap',
+    overflow: 'scroll',
```
MrOrz, line 39: If what we want is a horizontal scroll when there is not enough space, please use `overflowX: auto`.

`overflow: scroll`:
- introduces always-on scrolls, it's there even if no scroll is needed
- scrolls both directions, which is pretty rare in design

image

I suggest web developers on Mac should turn on "always show scroll bars" at all times so that we can empathize Windows users more. It may be not obvious for Mac users with hidden scrollbars, the [Windows users will always get ugly scroll bars for every `overflow: scroll` we write.](https://www.filamentgroup.com/lab/scrollbars/)

pages/article/[id].js
```diff
@@ -10,22 +13,99 @@ import useCurrentUser from 'lib/useCurrentUser';
...
+    [theme.breakpoints.up('md')]: {
+      flex: 3,
+      marginRight: 12,
+    },
+  },
+  aside: {
+    background: 'transparent',
+    [theme.breakpoints.up('md')]: {
+      padding: '16px 28px',
+      background: theme.palette.common.white,
```
Comment (MrOrz) [inline pages/article/[id].js:57]: seems that we are missing `flex: 1` here
image

pages/article/[id].js
```diff
@@ -10,22 +13,99 @@ import useCurrentUser from 'lib/useCurrentUser';
...
+    marginRight: 0,
+    [theme.breakpoints.up('md')]: {
+      flex: 3,
+      marginRight: 12,
+    },
+  },
+  aside: {
+    background: 'transparent',
+    [theme.breakpoints.up('md')]: {
+      padding: '16px 28px',
```
Comment (MrOrz) [inline pages/article/[id].js:56]: `.aside`'s padding and `.similarMessageContainer`'s horizonal padding are stacked together on desktop.
image

Mockup:
image

pages/article/[id].js
Comment (yanglin5689446) [reply pages/article/[id].js:337]: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=1307%3A285
the mobile version is horizontal scrolls

Comment (yanglin5689446) [reply pages/article/[id].js:337]: I'll change it to `overflow: auto` though, to hide scrollbar when user using windows

components/PlainList.js
```diff
@@ -0,0 +1,13 @@
+import { withStyles } from '@material-ui/core/styles';
+import cx from 'clsx';
+
+const PlainList = withStyles({
```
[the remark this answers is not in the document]
yanglin5689446: card added here
https://github.com/orgs/cofacts/projects/5#card-38799010

components/CreateReplyRequestForm/CreateReplyRequestForm.js
yanglin5689446: card added here
https://github.com/orgs/cofacts/projects/5#card-38799032

components/ArticleReply/index.js
yanglin5689446: I think it's because that the logic here is to display feebacks with user present:
```
{feedbacks
            .filter(({ vote, user }) => vote === 'UPVOTE' && user)
            .map(feedback => (
              <Feedback key={feedback.id} {...feedback} />
            ))}
```

MrOrz: Got it, make sense.
I am going to [revamp user handling](https://g0v.hackmd.io/ZcoUOX_-RQSkJyl5xz4_Zg), giving users from `localhost` a dedicated browser app ID on staging site, and show browser app users to all apps. Hopefully this will solve these weird issue during development.

components/NewReplySection/Desktop.js
```diff
@@ -105,13 +105,21 @@ const Desktop = ({
       )}
       {selectedTab === 1 && (
         <>
+          <Box
+            display="flex"
+            alignItems="center"
+            justifyContent="space-between"
```
MrOrz, line 111: [Trivial] Since all its child have `flex-grow`, we don't need `justifyContent` here

components/RelatedReplies.js
```diff
@@ -81,7 +94,16 @@ function RelatedReplyItem({ article, reply, onConnect, disabled }) {
   return (
     <li className={classes.root}>
       <Link href="/article/[id]" as={`/article/${article.id}`}>
-        <a className={classes.link}>相關訊息</a>
+        <a className={classes.link}>
```
MrOrz, line 93: Let's add `target="_blank"` here since the button says "in new tab"

pages/article/[id].js
```diff
@@ -338,7 +336,7 @@ function ArticlePage() {
             >
               {similarArticles.map(({ node }) => (
                 <div key={node.id} className={classes.similarMessageContainer}>
-                  <article className={classes.text}>{node.text}</article>
+                  <article className={classes.text}>{nl2br(node.text)}</article>
```
Comment (MrOrz) [inline pages/article/[id].js:346]: The desktop now looks good in line clamp :)

image

However, given the limited lines to display in this section, I think we don't need to add `nl2br`, so that we can reveal more text to the user.

Comment (MrOrz) [reply pages/article/[id].js:346]: > The desktop now looks good in line clamp :)
> 
> However, given the limited lines to display in this section, I think we don't need to add `nl2br`, so that we can reveal more text to the user.
> 
> image

pages/article/[id].js
```diff
@@ -338,7 +336,7 @@ function ArticlePage() {
             >
               {similarArticles.map(({ node }) => (
                 <div key={node.id} className={classes.similarMessageContainer}>
```
Comment (MrOrz) [inline pages/article/[id].js:345]: Clicking each item should go to article page. Forgot to add `<Link>` :P

components/ArticleReply/ReplyShare.js
```diff
@@ -0,0 +1,123 @@
...
+        type="button"
+        className={cx(classes.button, classes.desktop)}
+        onClick={openMenu}
+      >
+        {t`Share`}
+      </button>
+      <button
+        type="button"
+        className={cx(classes.button, classes.mobile)}
+        onClick={handleShare}
```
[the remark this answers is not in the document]
MrOrz: On desktop browser with a narrow screen, we get an inactive button:

share-no-use

Root cause is as described above -- we only need one button that invokes `navigator.share` if it's available and shows dropdown otherwise.

pages/article/[id].js
```diff
@@ -139,104 +228,141 @@ function ArticlePage() {
...
+          <h4>{t`Similar messages`}</h4>
+          {similarArticles.length ? (
+            <Box
+              display="flex"
+              flexDirection={{ xs: 'row', md: 'column' }}
+              overflow="auto"
+            >
+              {similarArticles.map(({ node }) => (
+                <Link key={node.id} href={`/article/${node.id}`}>
+                  <div className={classes.similarMessageContainer}>
```
Comment (MrOrz) [inline pages/article/[id].js:348]: Next.js [suggests](https://nextjs.org/docs/api-reference/next/link) using `<a>` tag for `<Link>`'s children. I think we can directly change `.similarMessageContainer` to use `a` tag instead of `div`.

Currently it do navigates user if they click the link, but if `<Link>` is combined with `<a>`, it allows user to cmd-click to open the link in new tab. We won't need `cursor: pointer` on the element as well, since `<a>` already turns cursor into pointer. (But we may need `text-decoration: none` to cancel out `<a>`'s default style, though.)

components/ArticleReply/index.js
```diff
@@ -162,39 +161,17 @@ const ArticleReply = React.memo(
...
-          <span key="editor">
-            {jt`${articleReplyAuthorName} uses ${originalAuthorsReply} to`}
-          </span>
-        );
-      }
-
-      return articleReplyAuthorName;
-    }, [articleReplyAuthor, replyAuthor]);
+        ) || t`Someone`,
+      [articleReplyAuthor, replyAuthor]
```
MrOrz, line 173: Trivial: `replyAuthor` seems not needed here

components/NewReplySection/index.js
```diff
@@ -0,0 +1,165 @@
...
+    const [connectReply, { loading: connectingReply }] = useMutation(
+      CONNECT_REPLY,
+      {
+        refetchQueries: ['LoadArticlePage'],
+        awaitRefetchQueries: true,
+        onCompleted() {
+          onSubmissionComplete(); // Notify upper component of submission
+          handlers.clear();
+          onClose();
+          setFlashMessage(t`Your have attached the reply to this message.`);
```
MrOrz, line 98: This flash message is never shown because `onClose()` unmounts `Snackbar`. Suggest moving snackbar to in `article/[id].js`.

PR #250, opened 2020-05-06
New Pagination component UI

yanglin5689446:
Change pagination to infinity load


components/LoadMore.js
```diff
@@ -0,0 +1,58 @@
...
+    return null;
+  }
+
+  const lastCursorOfPage =
+    edges.length && edges[edges.length - 1] && edges[edges.length - 1].cursor;
+
+  const loadMore = () => {
+    if (lastCursorOfPage !== lastCursor) {
+      setCursor(lastCursorOfPage);
+      load();
```
MrOrz, line 40: If we leverage `loadMore` and get cursor from last edge directly in `ArticlePageLayout`, we can leave the pagination logic entirely in where pagination happens, reduce the props passed to `LoadMore`.

yanglin5689446: I eventually remove the LoadMore component and leverage the whole component to `AarticlePageLayout`, because with only `onClick` &  `loading` the LoadMore component seems similar to regular button, the actual logics are mostly  move to `ArticlePageLayout`, it's a bit strange to have a Component call `LoadMore` while the actual load-more logics are written in upper component.

components/ArticlePageLayout.js
```diff
@@ -88,6 +88,7 @@ const useStyles = makeStyles(theme => ({
   },
   articleList: {
     padding: 0,
+    marginBottom: -12,
```
MrOrz, line 92: Sorry, I forgot to consider the case when there is no "next" button:

image

(It's fine in mobile, though; it's just that floating action buttons are covering the last item)

PR #253, opened 2020-05-17
Traditional Chinese translation for list pages

MrOrz:
Translate to zh_TW of list pages, before pushing to production.

Translations referring to the [spec](https://g0v.hackmd.io/@NFi0czulSemxCM8RNSlz8Q/HJ8xT3QVU) and [mockup](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=0%3A1).

image

image

image

image

image

### DIFF
--- a/i18n/zh_TW.po
+++ b/i18n/zh_TW.po
@@ -833,11 +833,11 @@ msgstr "看理由"
 #: components/ReplyFeedback.js:349
 #, javascript-format
 msgid "Helpful ${ positiveFeedbackCount }"
-msgstr "有幫助"
+msgstr "有幫助 ${ positiveFeedbackCount }"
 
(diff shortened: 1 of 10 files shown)


i18n/zh_TW.po
```diff
@@ -841,22 +720,266 @@ msgstr ""
...
+
+#: components/TimeRange.js:65
+msgid "In 1 Month"
+msgstr "一個月內"
+
+#: components/TimeRange.js:163
+msgid "to"
+msgstr "到"
+
+#: components/ArticleCategories/AddCategoryDialog.js:86
```
MrOrz, line 871: These are temporary components that is going to be replaced, so I don't translate these.

PR #256, opened 2020-05-28
Feature/search page

yanglin5689446:
Update search page, make replies search works.
mockup: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=432%3A2727

### DIFF
--- a/components/ReplySearchItem.js
+++ b/components/ReplySearchItem.js
@@ -66,12 +66,15 @@ const useStyles = makeStyles(theme => ({
     display: 'none',
     position: 'absolute',
     right: theme.spacing(2),
+    top: 0,
+    bottom: 0,
     '& > svg': {
       verticalAlign: 'middle',
       marginLeft: theme.spacing(1),
     },
     [theme.breakpoints.up('md')]: {
-      display: 'inline',
+      display: 'flex',
+      alignItems: 'center',
     },
   },
   article: {
@@ -104,7 +107,7 @@ export default function ReplySearchItem({
     setOpen(false);
   };
 
-  const replyCount = articleReplies.length;
+  const replyCount = articleReplies.length - 1;
 
   const {
     article,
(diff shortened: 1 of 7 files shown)


components/ReplySearchItem.js
```diff
@@ -0,0 +1,203 @@
...
+          replyId={reply.id}
+          reply={reply}
+          feedbacks={feedbacks}
+          positiveFeedbackCount={positiveFeedbackCount}
+          negativeFeedbackCount={negativeFeedbackCount}
+          ownVote={ownVote}
+          query={query}
+        />
+      </Box>
+      <button
```
MrOrz, line 141: This button should show only when `articleReplies.length` > 1. If there is only one `articleReply`, it means that the reply is only used in one aritcle, and there is no "other messages" to show.

The second item in the mockup shows the variant of `ReplySearchItem` with `articleReply.length` === 1.
image

components/ReplySearchItem.js
```diff
@@ -0,0 +1,203 @@
...
+        <span className={classes.view}>
+          {t`view`}
+          <VisibilityIcon />
+        </span>
+      </button>
+      <Dialog open={open} onClose={handleClose}>
+        <DialogTitle>{t`This reply is used in following messages`}</DialogTitle>
+        <DialogContent>
+          <PlainList>
+            {articleReplies.map(({ article }) => (
```
MrOrz, line 160: Since this button is supposed to show "other" messages, we should not repeat the article (first original article) shown outside of the dialog.

Please filter out the article that is already shown above.

螢幕快照 2020-06-01 下午2 09 28

MrOrz:
I forgot to mention that the search result page currently have a mysterious horizontal scroll. Please find what caused this scroll and eliminate it.

search-page-horizontal

yanglin5689446:
> I forgot to mention that the search result page currently have a mysterious horizontal scroll. Please find what caused this scroll and eliminate it.
> 
> search-page-horizontal

I can't reproduce this issue, can you provide more information about the platform or testcase?

MrOrz:
> > I forgot to mention that the search result page currently have a mysterious horizontal scroll. Please find what caused this scroll and eliminate it.
> > search-page-horizontal
> 
> I can't reproduce this issue, can you provide more information about the platform or testcase?

It's visible if the user has "always show scroll" turned on.
image

On environments that always has scroll bars (such as Google Chrome 83 under Windows 10), the horizontal bar also exists:

image

I have located the root cause -- it's due to `width: 100vw` and `min-width: 100vw` set on jumbotron.

- `100vw` measures the viewport size with scroll bar included.
- Whenever there is something that is outside of right & bottom edge of viewport, scrollbars are generated.
- Therefore, on environments that scroll bars do take space, using `100vw` on a page that is long enough to generate vertical scrollbar will generate a horizontal scrollbar on body.

Replacing width setup with `right: 0` fixes the issue.

image

PR #258, opened 2020-05-29
Enable react hook lint rules

MrOrz:
This PR enables react-hooks eslint rules and fixes all errors.

All violation of `react-hooks/exhausive-deps` now triggers build failure and requires explicitly disabling it using line comments (also need to include a good reason in comment!).

Current violations are fixed using the following rules:

-  `useCallback`:
  1. curried callbacks (`item => () => {}`) should not put in `useCallback` because when it's used in event handlers, curry function is invoked and a new copy of handler is always returned. There is no use putting `useCallback` over curried callbacks.
  2. event handlers depending on other function that is not memoized: its dependency will change every render. I remove its `useCallback` in this case.
  3. otherwise, add dependency to array. 
- `useEffect`:
  1. If it's meant to be a `componentDidMount`, I disable the rule for that line.
  2. If it's meant to be invoked on dependency change, I add the remaining dependencies.

**Each code change in this PR may introduce hard-to-find bugs. Please inspect carefully and put a comment whenever in doubt.**

### DIFF
--- a/components/TimeRange.js
+++ b/components/TimeRange.js
@@ -1,4 +1,4 @@
-import { useRef, useState, useEffect, useCallback, forwardRef } from 'react';
+import { useRef, useState, useEffect, forwardRef } from 'react';
 import DateRangeIcon from '@material-ui/icons/DateRange';
 import { makeStyles } from '@material-ui/core/styles';
 import { ButtonGroup, Button, Menu, MenuItem } from '@material-ui/core';
@@ -66,6 +66,18 @@ const options = [
   { value: 'custom', label: c('Time range dropdown').t`Custom` },
 ];
 
+/**
+ * @param {{GTE: string?, LTE: string?}} range
+ * @return {string} one of options' value
+ */
+function getSelectedFromRange(range) {
+  if (!range) return 'all';
+  if (range.LTE) return 'custom';
+
+  const option = options.find(o => o.value === range.GTE);
+  return option ? option.value : 'custom';
+}
+
 /*
   material ui actually passes down some non-DOM props to children,
   and this cause some warning on runtime.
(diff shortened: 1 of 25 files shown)


components/TimeRange.js
```diff
@@ -126,7 +123,7 @@ function TimeRange({ onChange = () => null, range }) {
         }
       }
     }
-  }, [range]);
+  }, [range, customValue]);
```
MrOrz, line 126: @yanglin5689446 This is causing lint error because of the use of `setAndUpdateValue`, which is updated every `render`. The logic in this hook is a bit complicated and its side effect (invoking `setAndUpdateValue`) is costly.

Would you take a look at this and determine an appropriate dependency list here?

yanglin5689446: I think it's safe to include only `range` as dependency, because this effect should only run when props passed from parent update, only when this happen the effect should be run and bind to new closure, though I'm not pretty sure if it works as I expected 😅

yanglin5689446: with `customValue` included in deps list, the state in component will always override the props passed from parent so the props update from parent never applied(correct me if I'm wrong)

MrOrz: Got it. In the latest change, I did not put `customValue` in the dep list.

However, the use of `setAndUpdateValue` still leaves opportunity that this `useEffect` function can invoke old `onChange` prop from some very old closure. According to react doc, it is [not safe to omit functions from the list of dependencies](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies).

I have re-written all logic in `useEffect` using functional updates of states, as suggested in [React documentation](https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often).  Hopefully it works all the same.

PR #261, opened 2020-06-08
Improve article category UI

yanglin5689446:
New design for article page category system, [mockup].(https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=912%3A303)
some behaviors are still weird, just pop this version out to collect more feedbacks.

### DIFF
--- a/components/ArticleCategories/AddCategoryDialog.js
+++ b/components/ArticleCategories/AddCategoryDialog.js
@@ -1,47 +1,12 @@
 import { t } from 'ttag';
 import gql from 'graphql-tag';
+import { Dialog, DialogTitle, DialogContent } from '@material-ui/core';
+
 import { useMutation } from '@apollo/react-hooks';
-import Link from 'next/link';
-
-import Grid from '@material-ui/core/Grid';
-import DialogTitle from '@material-ui/core/DialogTitle';
-import Dialog from '@material-ui/core/Dialog';
-import DialogContent from '@material-ui/core/DialogContent';
-import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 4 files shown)


components/ArticleCategories/AddCategoryDialog.js
```diff
@@ -155,26 +87,23 @@ function AddCategoryDialog({
...
-              className={classes.gridItem}
-            >
-              <CategoryOption
-                category={category}
-                onAdd={handleAdd}
-                selected={isSelectedMap[category.id]}
-                disabled={loading}
-              />
-            </Grid>
+              articleId={articleId}
```
MrOrz, line 98: Since addition, deletion, upvote and downvote all happens within the option, I think it is more logical if we handle these logic together -- we should handle all these actions either within `<CategoryOption>`, or handling them all in `<AddCategoryDialog>` and put these callback prop on `<CategoryOption>`.

As we pass `articleId` down to `CategoryOption` to handle its deletion, upvote and downvote, I think its good to move addition logic down to `CategoryOption` as well.

components/ArticleCategories/CategoryOption.js
```diff
@@ -326,7 +363,7 @@ function CategoryOption({
         <DownVoteDialog
           articleId={articleId}
           categoryId={category.id}
-          onClose={() => showDownVoteDialog(false)}
+          onClose={() => setDownVoteDialogShow(false)}
```
MrOrz, line 366: Hmm I missed this in last review? XD

PR #263, opened 2020-06-13
Doc/storybook components

yanglin5689446:
Add stories to some components.
also remove unused stylesheet file in 220be59


yanglin5689446:
> LGTM!
> 
> IMO some adjustments can be made on `ExpandableText`. Currently the "閱讀全文" does not align with the text line:
> image
> 
> We can do the following to fix that:
> 
> * set `button`'s `padding` to 0, `vertical-align` to `bottom`
>   
>   * This removes padding around native `<button>` and makes sure surrounding `<div>` have the same height with `<button>`
> * set the `button` element's `font-size` and `line-height` to `inherit`
>   
>   * Ensures the height of the button will be the same height as its surroundings, so that the gradient color can cover exactly 1 line.
>   * [Replaced elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element) like `button` don't automatically inherit styles from document.
> 
> After adjustment:
> image

make sense. will fix it in fine tuning branch.

PR #264, opened 2020-06-18
Enhance/fine tuning

yanglin5689446:
- Some UI tuning
- User avatar with label
- Add float edit button to article details page


components/AppLayout/GlobalSearch.js
```diff
@@ -123,14 +138,14 @@ function GlobalSearch({ onIconClick }) {
             <div className={classes.resultEntry} onClick={navigate('messages')}>
               <div className={classes.iconWithText}>
```
MrOrz, line 139: Removing `iconWithText` div and just use `resultEntry` as flex container should yield same result with simpler structure.

image

Also, we may need to add `flex-shrink: 0` to `.right`.

components/AppLayout/GlobalSearch.js
```diff
@@ -60,6 +66,14 @@ const useStyles = makeStyles(theme => ({
     display: 'flex',
     alignItems: 'center',
   },
+  ellipsisOnMobile: {
+    [theme.breakpoints.down('md')]: {
+      maxWidth: 120,
```
MrOrz, line 71: 1. Seems that ellipsis is also needed on desktop 
image
2. In order to trigger overflow, we should set `min-width: 0` to override the box's [default min-sizing behavior](https://css-tricks.com/flexbox-truncated-text/) (which enforces the flex children to have same width as content, thus no overflow could happen). In this case, we won't need to set a fixed `max-width`, and the style should work on different screen widths.
image

pages/article/[id].js
```diff
@@ -39,6 +39,38 @@ const useStyles = makeStyles(theme => ({
...
+      padding: '10px 16px',
+      fontSize: 12,
+    },
+    '&:after': {
+      position: 'absolute',
+      top: 0,
+      right: 0,
+      borderRight: `20px solid ${theme.palette.common.white}`,
+      borderBottom: `20px solid ${theme.palette.primary[500]}`,
+      borderTop: `20px solid ${theme.palette.primary[500]}`,
```
Comment (MrOrz) [inline pages/article/[id].js:59]: Unfortunately the tails are not aligned in narrower screen:
image
(Actually the one in the menu is not aligned either.)

Since it's difficult calculating these triangles, I would suggest using svg instead.
It's shape is so simple that we can write the svg [by hand](https://www.oxxostudio.tw/articles/201406/svg-04-path-1.html):
```
<svg viewBox="0 0 1 2">
  <path d="M0 0 H1 L0 1 L1 2 H0 Z" />
</svg>
```
And we get a ribbon tail that we can resize and recolor using CSS.

pages/article/[id].js
```diff
@@ -263,15 +296,15 @@ function ArticlePage() {
           <Box
             className={classes.card}
             position="relative"
-            px={{ xs: '12px', md: '19px' }}
-            py={{ xs: '13px', md: '21px' }}
+            pb={{ xs: '13px', md: '21px' }}
```
Comment (MrOrz) [inline pages/article/[id].js:299]: Mysterious misalignment
螢幕快照 2020-06-21 下午5 17 43

components/AppLayout/GlobalSearch.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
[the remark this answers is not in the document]
yanglin5689446: I think it's accidentally lol

components/ReplyFeedback.js
```diff
@@ -291,28 +280,22 @@ function ReplyFeedback({
...
-        className={cx(classes.vote, ownVote === 'DOWNVOTE' && classes.voted)}
-        type="button"
-        onClick={e => openVotePopover(e, 'DOWNVOTE')}
-        data-ga="Downvote"
-      >
-        <ThumbDownIcon className={classes.outlinedIcon} />
-      </button>
       <div className={classes.buttonGroup} data-ga="Number display">
-        <button className={classes.vote} type="button">
+        <button
```
MrOrz, line 284: The height of the buttons are not even. 
image

When hovering on the button, the border highlight is scattered:
image

The same goes to the button groups in `CreateReplyRequestForm`. They have different heights (36px vs 38px).
imba-buttons

I would suggest we use [outlined button groups](https://material-ui.com/components/button-group/#sizes-and-colors) for all the connected buttons buttons, and [`<Button variant="contained" color="primary" disableElevation>`](https://material-ui.com/components/buttons/#contained-buttons) for "Reply to this message" button.

In this way we get even-height buttons, more sensible border highlight when hovering on the button, and a consistent feedback when clicking on the buttons.

components/ReplyFeedback.js
```diff
@@ -291,28 +280,22 @@ function ReplyFeedback({
...
+        >
           {positiveFeedbackCount}
           <ThumbUpIcon className={classes.thumbIcon} />
         </button>
-        <button className={classes.vote} type="button">
+        <button
+          className={cx(classes.vote, ownVote === 'DOWNVOTE' && classes.voted)}
+          type="button"
+          onClick={e => openVotePopover(e, 'DOWNVOTE')}
+          data-ga="Downvote"
```
MrOrz, line 297: Since we block users from upvoting / downvoting their own reply, I think we should directly disable upvote / downvote button here, instead of displaying a popover that cannot submit.

components/Filters.js
```diff
@@ -83,10 +91,11 @@ const Option = withStyles(theme => ({
     padding: '4px 10px',
     margin: '4px 2px',
     cursor: 'pointer',
+    border: ({ chip, selected }) =>
+      chip || selected ? `1px solid ${theme.palette.secondary[100]}` : 'none',
```
MrOrz, line 95: Selected categories should also have borders

Figma:
image


Now:
image

components/CreateReplyRequestForm/CreateReplyRequestForm.js
```diff
@@ -149,6 +224,19 @@ const CreateReplyRequestForm = React.memo(
...
+    // event scroll listener
+    useEffect(() => {
+      function handler() {
+        if (!inViewport(buttonRef.current) && !showFloatButton) {
+          setShowFloatButton(true);
+        } else if (inViewport(buttonRef.current) && showFloatButton) {
+          setShowFloatButton(false);
+        }
+      }
+      window.addEventListener('scroll', handler);
```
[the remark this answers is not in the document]
yanglin5689446: I found this API, but the browser support is [fairly poor](https://caniuse.com/#search=intersection%20observer) as for now

pages/article/[id].js
```diff
@@ -284,40 +317,41 @@ function ArticlePage() {
...
+                    articleId={article.id}
+                    replyRequest={replyRequest}
+                    isArticleCreator={idx === 0}
+                  />
+                ))}
+                <CreateReplyRequestForm
+                  requestedForReply={article.requestedForReply}
+                  articleId={article.id}
+                  onNewReplyButtonClick={() => {
+                    setShowForm(true);
```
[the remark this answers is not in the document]
Comment (yanglin5689446) [reply pages/article/[id].js:350]: I add the scrollIntoView feature in `useEffect` on Desktop editor so that when either float or regular button is clicked the editor scroll into view, and avoid deep ref pass-down issue.

components/CreateReplyRequestForm/CreateReplyRequestForm.js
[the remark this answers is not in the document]
MrOrz: There are 2 versions of `IntersectionObserver`. The one that calculates intersection with viewport is the first version, which has pretty descent support: https://caniuse.com/#feat=intersectionobserver

PR #269, opened 2020-06-28
Feature/reply editor toobar

yanglin5689446:
Add some features to reply editor
- list style support
- emoji picker
- extend from other reply


components/NewReplySection/ReplyForm/ReasonEditor.js
```diff
@@ -80,74 +154,173 @@ const useStyles = makeStyles(theme => ({
...
+    const raw = element.value;
+    element.value = raw.slice(0, index) + emoji.native + raw.slice(index);
+    // mock a onchange event
+    onChange({ target: { value: element.value } });
+    setShowEmojiPicker(false);
+  };
+
+  const handleConnect = reply => {
+    let replyReference = reply.text + '\n';
+    // reply.hyperlinks.forEach(({ title, url }) => replyReference += )
```
MrOrz, line 212: 1. `replyReference` does not look like a variable that is really needed. If we need this variable, please use `const` and replace `reply.text + '\n'` with `replyReference` instead.
2. Please remove unused comment.
3. Seems that we did not copy the reply reference to `ReferenceInput` yet?

components/NewReplySection/ReplyForm/ReasonEditor.js
```diff
@@ -80,74 +155,178 @@ const useStyles = makeStyles(theme => ({
...
+    const index = element.selectionStart;
+    const raw = element.value;
+    element.value = raw.slice(0, index) + emoji.native + raw.slice(index);
+    // mock a onchange event
+    onChange({ target: { value: element.value } });
+    setShowEmojiPicker(false);
+  };
+
+  const handleConnect = reply => {
+    const hyperlinks = reply.hyperlinks.map(
```
MrOrz, line 213: Thanks for adding `reference`.

I think we should use `reply.reference` directly instead of composing from `reply.hyperlinks`. Sometimes the author of the reply may [put some info in reply's reference field](https://old.cofacts.org/reply/ORLVmHEBrhVJn3LNJnmn), such as adding section title between hyperlinks, or provide description to URLs that is more concise to hyperlink titles.

PR #272, opened 2020-07-03
hotfix: add category bug

yanglin5689446:
Add user id field to CreateArticleCategory

### DIFF
--- a/components/ArticleCategories/ArticleCategory.js
+++ b/components/ArticleCategories/ArticleCategory.js
@@ -21,6 +21,7 @@ const ArticleCategoryData = gql`
     negativeFeedbackCount
     ownVote
     canUpdateStatus
+    status
   }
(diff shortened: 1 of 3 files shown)


MrOrz:
Still can't delete category. The category first seems to delete, but re-appear after refresh.

I am seeing this warning in console. Not sure if it is relevant.

> Warning: fragment with name ArticleWithCategories already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names

I found that
- there is no `status` exist in fetched `articleCategories` in apollo cache.
- when querying article's `articleCategories` we did not specify `(status: NORMAL)` in GraphQL.

I guess that's probably why all `articleCategories` are fetched and displayed?

PR #297, opened 2020-07-25
Add search icon, placeholder, and enablef height adjustment

AJua:
Done
搜尋框加上「參考過往回應」的 placeholder
使理由與來源的 input 能垂直拉大
增加搜尋圖示提供搜尋功能，而不是只能用 Enter

TODOs
使理由與來源的 input 水平方向則固定
搜尋時加上 loading indicator
修正 iOS 點按「Add this reply to my reply」送出的問題

### DIFF
--- a/components/NewReplySection/ReplyForm/ReferenceInput.js
+++ b/components/NewReplySection/ReplyForm/ReferenceInput.js
@@ -13,7 +13,6 @@ const useStyles = makeStyles(theme => ({
     },
   },
   textarea: {
-    flex: 1,
     width: '100%',
     borderRadius: 8,
     border: 'none',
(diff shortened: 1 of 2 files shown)


components/NewReplySection/ReplyForm/ReferenceInput.js
```diff
@@ -13,7 +13,6 @@ const useStyles = makeStyles(theme => ({
     },
   },
   textarea: {
-    flex: 1,
```
MrOrz, line 16: Thank you for locating the root cause!

After trying the UI out, I found that `flex: 1` is actually required on mobile to enlarge the textarea.

I suggest leaving `flex: 1` as-is and adding the following in `[theme.breakpoints.up('md')]`: 
```
      flex: 'none', // Disable flex so that user can enlarge textarea by themselves
      minHeight: 144
```

AJua: Thanks for your review. 
I had applied the suggestions.

Issue #280, opened 2020-07-11
iOS users cannot perform search


MrOrz:
After iOS users finish input in top search bar, the search bar would disappear.

Here is a simulation using Safari:
![ios search](https://user-images.githubusercontent.com/108608/87223969-cf239780-c3b3-11ea-801a-95ac7359087c.gif)

On real iOS devices, as long as soft keyboard is dismissed, the search bar disappears.

Replacing the current on-blur mechanism with Material-UI [Click-away listener](https://material-ui.com/components/click-away-listener/#click-away-listener) should work.

bestian:
Hi @MrOrz, this looks already fixed on `master` by [`d7aa3dd`](https://github.com/cofacts/rumors-site/commit/d7aa3dda75dd3e83ca577e6a9a387b0422b9f86f) ("Prevent search input from being hided", merged Jul 2020, together with `e7d103c` / `3753a58`).

`components/AppLayout/GlobalSearch.js` now collapses the mobile search only through MUI `ClickAwayListener`'s `onClickAway` — there is no on-blur collapse anymore, so dismissing the iOS soft keyboard no longer hides the search bar. The installed MUI v4 `ClickAwayListener` also registers `onTouchEnd` in addition to `onClick`, so touch-away works on iOS.

We also added a regression test in our fork (blur does not collapse / click-away collapses): https://github.com/bestian/rumors-site/tree/fix/issue-280-ios-search — feel free to close this issue. Thanks!

MrOrz:
Thanks @bestian ! We plan to rewrite the whole site, it's nice to know these old issues are already gone.

PR #299, opened 2020-07-25
Prevent search input from being hided

AJua:
Fix issue #280

### DIFF
--- a/components/AppLayout/AppHeader.js
+++ b/components/AppLayout/AppHeader.js
@@ -232,10 +232,7 @@ function AppHeader({
             <Links classes={classes} unsolvedCount={unsolvedCount} />
           </Box>
         </div>
-        <GlobalSearch
-          displayLogo={() => setDisplayLogo(true)}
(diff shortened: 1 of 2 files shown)


components/AppLayout/AppHeader.js
```diff
@@ -232,7 +232,7 @@ function AppHeader({
             <Links classes={classes} unsolvedCount={unsolvedCount} />
           </Box>
         </div>
-        <GlobalSearch onIconClick={() => setDisplayLogo(false)} />
```
MrOrz, line 235: Personally I would prefer `onExpanded` or `onToggle` because
this is the scenario that a child component component notifies 

```js
 <GlobalSearch onExpanded={expanded => setDisplayLogo(!expanded)} />
```

AJua: I've refactor the relation between the parent and the child component so that parent listen to child's change rather then the child change the parent directly

PR #301, opened 2020-07-29
Introduce <Infos> and <TimeInfo> for as a flexible info display

MrOrz:
This PR implements `<Infos>` and `<TimeInfo>` and reuse them in `<ArticleInfo>` and `<ReplyInfo>` temporarily.

Actually, in the mockup the info lines are very diverse and has different info to display in each place.

In the future,  `<ArticleInfo>` and `<ReplyInfo>` will be removed, and `<Infos>` and `<TimeInfo>` will be directly used in article items instead.

## Storybook

info optional
TimeINfo

## Reply list
image

## Article detail
image

### DIFF
--- a/components/ReplyInfo.js
+++ b/components/ReplyInfo.js
@@ -1,10 +1,7 @@
 import gql from 'graphql-tag';
 import { t, ngettext, msgid } from 'ttag';
 import Link from 'next/link';
-import isValid from 'date-fns/isValid';
-import { format, formatDistanceToNow } from 'lib/dateWithLocale';
-import Tooltip from './Tooltip';
-import Infos from './Infos';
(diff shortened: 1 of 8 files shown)


components/ReplyInfo.js
```diff
@@ -1,74 +1,34 @@
...
-      {user?.name && (
-        <span
-          className={classes.info}
-        >{t`originally written by ${user.name}`}</span>
-      )}
+    <Infos>
+      <TimeInfo time={createdAt}>
+        {timeAgoStr => (
+          <Link href="/reply/[id]" as={`/reply/${reply.id}`}>
+            <a>{t`replied ${timeAgoStr} ago`}</a>
```
MrOrz, line 18: `color` should be `inherit`

PR #305, opened 2020-08-04
RSS UI

nonumpa:
figma
https://www.figma.com/file/YikWPPauMnukDH0U2Nq4YS/Components?node-id=1%3A310

desktop-zh

mobile-en

### Subscribe through IFTTT
image

Not Ready:
1. Icon (currently download and search from google image)
2. Copy rss feed button is better to popup a tip 'copied to clipboard'.


components/Subscribe/IFTTTItem.js
```diff
@@ -0,0 +1,155 @@
...
+import CloseIcon from '@material-ui/icons/Close';
+import IconButton from '@material-ui/core/IconButton';
+
+import { makeStyles } from '@material-ui/core/styles';
+import { ButtonIcon } from './FeedDisplayComponents';
+
+import CopyButton from '../CopyButton';
+
+const useStyles = makeStyles(theme => ({
+  paper: {
```
MrOrz, line 17: If we use [`Dialog`](https://material-ui.com/components/dialogs/) instead of its underlying component [`Modal`](https://material-ui.com/components/modal/), we would not need to define paper style and positioning here. 

Dialogs are not necessary used with `DialogTitle`, `DialogContent` etc, it can be used alone, as shown by `ListPageControls`' `Filters`.

components/Subscribe/IFTTTItem.js
```diff
@@ -0,0 +1,155 @@
...
+          className="iframe"
+          width="100%"
+          height="100%"
+          src={`https://www.youtube.com/embed/${tutorialYoutubeId}`}
+          frameBorder="0"
+          allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
+          allowFullScreen
+        ></iframe>
+      </div>
+      <table className={classes.table}>
```
MrOrz, line 81: It's not recommended to use `<table>` for something that is not a data table.
I suggest using flexbox for this particular UI, which should yield much simpler DOM and CSS style.

components/ArticleReply/ReplyShare.js
```diff
@@ -53,7 +53,7 @@ const ReplyShare = ({ copyText }) => {
 
   const onSuccess = text => {
     setMessage(text);
-    setTimeout(() => setStatus('SUCCESS'), 0);
+    setTimeout(() => setStatus(SUCCESS), 0);
```
LucienLee, line 56: It seems that doesn't need to set success at next tick

components/Subscribe/IFTTTItem.js
LucienLee: agree

MrOrz (approved):
LGTM! Thanks for the fix. Let's merge this :D

PR #311, opened 2020-08-09
ArticleReplyFeedbackControl revamp

MrOrz:
This PR is a revamp to article-reply feedback component, including its popover:
image

This item is blocking #288, as the original `ReplyFeedback` props and data dependency is a bit hard to replicate across different pages.

## Refactors
- Rename `ReplyFeedback` to `ArticleReplyFeedbackControl`.
    - Split the button group component into `ArticleReplyFeedbackControl/ButtonGroupDisplay`
- Consolidate `ArticleReplyFeedbackControl`'s prop to `articleReply` and `reply`
    - This simplifies all occurrence of `<ArticleReplyFeedbackControl>`
    - Also simplifies GraphQL fragment definition for all components that uses `<ArticleReplyFeedbackControl>`, since it's clear that we need to attach fragment for the two prop `articleReply` and `reply`.
      -  updates `components/ArticleReply`
      - `ReplyItem`'s props are also updated to use `articleReply` instead of separate fields
    - We differentiate `reply` from `articleReply` because they are different entities, and we do not necessarily get reply from `articleReply`
- Add data requirements (GraphQL fragments) to `ArticleReplyFeedbackControl.fragments`
    - Also fixes the GraphQL fragments of all components that uses `<ArticleReplyFeedbackControl>`, and the components' ancestors.
- Move icons in ArticleReplyFeedbackControl to `components/icons` and add storybook
    - Will move other icons inside as we refactor other parts of the site
    - 
image

## TODO in future PRs
- Further split reason input popover & other's reason display popover into different display components 
    - They are `ReasonInput` and `ReasonsDisplay` respectively, both empty in this PR and will be populated in future PRs.
    - `ArticleReplyFeedbackControl.js` will contain these components and manage the display of these popovers, while all display logic & data loading logic delegated to these display components.
- Loads feedbacks only when "See Reasons" button is clicked
    - Leverage `ListArticleReplyFeedbacks` API
- Rewrite mutations to update `articleReply` instance in Apollo cache
    - Also provide `onVoteComplete` callback to allow individual pages to perform other actions (such as refresh the order of whole list of article replies)


MrOrz:
Merged due to inactivity

PR #312, opened 2020-08-09
Upgrade to Apollo Client 3

MrOrz:
Apollo Client 3 stable is [released](https://www.apollographql.com/blog/announcing-the-release-of-apollo-client-3-0/).

This PR [migrates Apollo Client to 3](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/) and [next-apollo to latest](https://github.com/adamsoffer/next-apollo/blob/master/CHANGELOG.md).

`dataIdFromObject` we are currently using is deprecated in Apollo Client 3; however, in this PR I leave this intact because there are already too much code change.

Other changes include:
- Make eslint-plugin-react happy by explicitly telling it to detect React version in eslintrc
- Add [client name](https://www.apollographql.com/docs/apollo-server/monitoring/metrics/#identifying-distinct-clients) when creating `ApolloClient` instance to make apollo graph manager recognize requests from rumors-site.


pages/search.js
```diff
@@ -179,4 +179,4 @@ function SearchPage() {
   );
 }
 
-export default withData(SearchPage);
+export default withApollo()(SearchPage);
```
MrOrz, line 182: Leaving out `{ssr: true}` in `withApollo` here will cause next.js's  `useRouter` returning `{}` for `router.query` on server-side rendering & first render on client, because next.js will think the page [does not meet data fetching requirements](https://nextjs.org/docs/api-reference/next/router#router-object).

It's valid for search page because this page is not meant to be accessed directly nor by search engines.

MrOrz:
Merged due to inactivity.

PR #315, opened 2020-08-15
Initial implementation of LineChart

ztsai:
Merging this PR to feature branch instead of `dev` because it's still a work in progress, and blocked by designs.
However, I would like to check with you and see if the current implementation aligns with the current coding style/standard before moving further, as we have discussed a few times about how to integrate `d3` with `React`.

Screen Shot 2020-08-15 at 4 34 40 PM


components/LineChart.js
```diff
@@ -0,0 +1,235 @@
+import * as d3 from 'd3';
```
MrOrz, line 1: I think this would import the whole `d3`, including things that we don't use at all, such as [d3.selection](https://github.com/d3/d3-selection).

I suggest importing individual packages and functions (i.e. `import {scaleLinear} from 'd3-scale';`).

components/LineChart.stories.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+
+import LineChart from './LineChart';
+
+export default {
+  title: 'LineChart',
+  component: 'LineChart',
+};
+
+const data = [
```
MrOrz, line 10: I am curious how `LineChart` would behave if 
1. there are holes in time (i.e. skipped date)
2. contains data with less than 30 days

Is it going to be handled in `LineChart` or is it going to be handled in some wrapper component?

ztsai: I believe we discussed this briefly at moed23ct, the ultimate solution is to populate data for all missing dates for the last 30 days.
The plan is to move the dataset handling part inside `getChartData` to the parent component and we can handle missing data there as well.
I think this component should just plot the charts with given data and layout

ztsai:
Thanks a lot for the feedback!  I'm marking this as WIP and will hold off until we get confirmation on the design.

components/LineChart.stories.js
MrOrz: Got it, agree that we can just plot the data given to `<LineChart>` to keep it simple. Thanks for the clarification!

ztsai:
Screen Shot 2020-08-30 at 8 18 31 PM
Screen Shot 2020-08-30 at 8 18 52 PM
Screen Shot 2020-08-30 at 8 18 43 PM
Screen Shot 2020-08-30 at 8 21 08 PM

Storyshots.test.js
```diff
@@ -58,14 +58,13 @@ function removeMaterialUIInternals(json) {
 
   return json;
 }
-
+MockDate.set('2020-01-01');
```
MrOrz, line 61: Actually I am not sure if `test()` is called first, or if `MockDate.reset()` is called first. Depends on how `initStoryshots` calls `test()`.

Judging from `Info`'s snapshot change (`over 30 years` -> `about 31 years`) I suspect that the mock is not in effect.

ztsai: you're right, I didn't think it through lol
I've modified the test accordingly.

components/LineChart.js
```diff
@@ -0,0 +1,346 @@
...
+    '& .rightAxis': {
+      color: theme.palette.primary[900],
+      fontWeight: 500,
+    },
+    '& .chatbotLine': {
+      stroke: theme.palette.primary.main,
+    },
+  },
+  wrapper: {
+    display: 'inline-block',
```
MrOrz, line 38: This can be omitted because of the existence of `position: absolute`

components/LineChart.js
```diff
@@ -0,0 +1,346 @@
...
+                left: boxWidth / 2 - 5 + 'px',
+              }}
+            ></div>
+            <div className={`tooltip ${i < 15 ? 'left' : 'right'}`}>
+              <div className="tooltip-title">{formatDate(d.date, true)}</div>
+              <div className="tooltip-body">
+                <p className="tooltip-subtitle">{t`Web Visit`}</p>
+                <p className="tooltip-text">
+                  <span>{d.webVisit}</span>{' '}
+                  <span>{c('LineChart').t`times`}</span>
```
MrOrz, line 332: We can use [ngettext](https://ttag.js.org/docs/ngettext.html) to handle plural form.
```
c('LineChart').ngettext(msgid`${num} time`, `${num} times`, num)
```

components/TrendPlot.js
```diff
@@ -0,0 +1,164 @@
...
+  const { dataset, totalLineVisits, totalWebVisits } = populateChartData(data);
+  const totalVisits = totalWebVisits + totalLineVisits;
+
+  return (
+    <div className={`${classes.root}`}>
+      <Box display="flex" flexDirection="row" alignItems="center" p={1}>
+        <Box className={`${classes.label} plotLabel`}>{t`past 31 days`}</Box>
+        <Hidden xsDown>
+          <Box className={`${classes.label} webLabel `}>
+            {jt`Web Visit: ${totalWebVisits}`}
```
MrOrz, line 130: Since `totalWebVisits` is not storing JSX elements, we can directly use [`t`](https://ttag.js.org/docs/tag-gettext.html) here, instead of [`jt`](https://ttag.js.org/docs/jsx-gettext.html).

The same goes to occurrences of `jt` below.

MrOrz (commented):
The chart is looking pretty good!

I have added some comments regarding some details of CSS and i18n.

components/LineChart.js
```diff
@@ -73,30 +73,30 @@ const useStyles = makeStyles(theme => ({
...
+        margin: 0,
       },
 
       '&.left': {
-        marginLeft: '40px',
+        left: '40px',
       },
       '&.right': {
-        marginLeft: '-100px',
+        left: '-100px',
```
MrOrz, line 99: `right: 40px` should have similar effect, but using `right` could have no dependency on tooltip actual width.

MrOrz (changes requested):
I found that currently the red cursor will not align with the x-axis:
image

To make them aligned, there should be 11 x-axis ticks for 31 days. For example, for 9/6 the ticks should be:
> 8/7 8/10 8/13 8/16 8/19 8/22 8/25 8/28 8/31 9/3 9/6

MrOrz:
Recently I have rolled back apollo client to version 2. Please also rebase the PR and change apollo-client 3 dependencies accordingly. More specifically,

- `import { gql } from ...` should be `import gql from 'graphql-tags'`
- `import { useQuery } from ...` should be `import { useQuery } from '@apollo/react-hooks'

components/LineChart.js
```diff
@@ -0,0 +1,349 @@
...
+    return `${d.getFullYear()}/${dateString}`;
+  }
+  return dateString;
+};
+
+function plotTicks({ scale, x, y, text, roundFn, tickNum, gridline }) {
+  const ticks = getTicks(scale, roundFn, tickNum);
+  return ticks.map((tick, i) => (
+    <TickGroup
+      key={`${tick - 0}_${i}`}
```
MrOrz, line 211: What does this `-0` for?

MrOrz (approved):
LGTM now :) Thanks for the fix and let's :shipit:

components/LineChart.js
ztsai: `tick` could be a `number` or a `date`, `-0` remains the same if it's a number and converts `date` into epoch seconds otherwise.
it's not necessary, just makes the `key` look cleaner

MrOrz: I see. We could use unary `+` to convert things to number, it's shorter ;)

PR #320, opened 2020-08-26
Remove weird replyRequestCount setup

MrOrz:
Removes weird `replyRequestCount` setup because it is causing errors if an user is visiting Cofacts with legacy URL (with `replyRequestCount` in its URL, such as `https://cofacts.org/articles?q=ooo&replyRequestCount=1`)

Also, it does not make sense to set `filterObj` without any condition. The line overrides the `replyRequestCount` filter set by `FILTERS.ASKED_MANY_TIMES`, thus it must be removed for `ASKED_MANY_TIMES` to work properly.

### DIFF
--- a/components/ArticlePageLayout.js
+++ b/components/ArticlePageLayout.js
@@ -137,8 +137,6 @@ function urlQuery2Filter({
     return undefined;
   }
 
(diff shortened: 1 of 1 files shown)


nonumpa (approved):
LGTM!

PR #321, opened 2020-08-26
Track rss, and add utm_source to each url that generated from feed.js

### LINKED ISSUE #222: Tracking RSS usage
We should track RSS usage.

Technologies:
- https://swizec.com/blog/tracking-rss-readers-with-google-analytics/swizec/4130
- image pixel: https://developers.google.com/analytics/resources/concepts/gaConceptsTrackingOverview
- `utm_source`: https://support.google.com/urchin/answer/2633622?hl=en , https://support.google.com/analytics/answer/1033863
  - `utm_source` [should be](https://holini.com/utm-tags/#11) RSS reader name from [user-agent string](https://www.petefreitag.com/item/418.cfm) ; note that ifttt identifies itself as ["curl"](https://justinmontgomery.com/measure-ifttt-traffic-via-naked-curl-user-agent-string)
  - `utm_medium` be `rss`
  - [ignore the rest](https://holini.com/utm-parameters/#11)
- https://en.wikipedia.org/wiki/RSS_tracking
- Requesting end - send GA events on RSS feed request

nonumpa:
Fixes #222 
## Feature
1. Track subscriber number by sending event: 
Event Category: RSS
Event Action: Subscribe
Event Label: `Source`(Feedly, Feedrabbit, IFTTT)
Document title: [response header](https://www.petefreitag.com/item/418.cfm)
Document path: query variables

    ##### Feedly
    - Sends subscriber number when subscribing.
    - Will cache the feed, it triggers the event when first time fetches the feed, and then each regular fetching(maybe [once an hour](https://www.feedly.com/fetcher.html)).

    ##### Feedrabbit 
    - Sends subscriber number each regular fetching.
    - Triggers the event every time.

2. Track article page through google tag manager by adding `utm_source` 
`utm_source`: `Source`(Feedly, Feedrabbit, IFTTT)
`utm_medium`: RSS

## Fix 
1. Copy URL function shouldn't encode URL

## Snapshots

Page title: res.header (subscriber number), Page path: query variables

utm_source to article page

### DIFF
--- a/components/Subscribe/FeedDisplay.js
+++ b/components/Subscribe/FeedDisplay.js
@@ -120,7 +120,7 @@ function FeedDisplay({ listQueryVars }) {
           </ListItemLink>
           <ListItemLink
             href={`https://feedly.com/i/discover/sources/search/feed/${encodeURIComponent(
-              feedUrl + `&source=feedly`
+              feedUrl + `&source=Feedly`
             )}`}
             icon={feedlyIcon}
             onClick={handleClose}
@@ -130,7 +130,9 @@ function FeedDisplay({ listQueryVars }) {
 
           <ListItemCopy
             icon={rssIcon}
-            textToCopy={encodeURIComponent(feedUrl + `&source=others`)}
+            textToCopy={
+              feedUrl + `&source=others&utm_source=Others&utm_medium=Subcribe`
+            }
             onSuccess={() => copySuccess(t`Copied to clipboard!`)}
           >
             {t`Get RSS Feed Link`}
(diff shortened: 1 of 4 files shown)


pages/api/articles/[feed].js
```diff
@@ -184,12 +185,23 @@ async function articleFeedHandler(req, res) {
       });
     });
 
+    const visitor = ua(PUBLIC_GA_TRACKING_ID);
+    const params = {
+      ec: 'RSS',
+      ea: 'Subscribe',
```
Comment (MrOrz) [inline pages/api/articles/[feed].js:191]: I think we can use something like `Poll`, which better describes this action.

components/Subscribe/FeedDisplay.js
```diff
@@ -128,7 +130,9 @@ function FeedDisplay({ listQueryVars }) {
 
           <ListItemCopy
             icon={rssIcon}
-            textToCopy={feedUrl}
+            textToCopy={
+              feedUrl + `&source=others&utm_source=Others&utm_medium=Subcribe`
```
MrOrz, line 134: What would `&utm_source=Others&utm_medium=Subcribe` do?

MrOrz (approved):
Thanks for the update, it's looking good!

Some thoughts and questions appended.

components/Subscribe/FeedDisplay.js
nonumpa: It should be `&source=Others` only, I forgot to remove `&utm_source=Others&utm_medium=Subcribe` after testing.

pages/api/articles/[feed].js
Comment (nonumpa) [reply pages/api/articles/[feed].js:191]: 👌

MrOrz (approved):

PR #335, opened 2020-09-20
Level icon in Profile page

MrOrz:
Figma: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=1678%3A510

image
image


components/LevelIcon.js
```diff
@@ -0,0 +1,359 @@
...
+      viewBox={`0 0 ${viewBoxWidth} ${viewBoxHeight}`}
+      preserveAspectRatio="xMidYMid meet"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      style={{ ...style, color: COLOR[colorIdx] }}
+      {...(id ? { id: `${id}` } : {})}
+      {...props}
+    >
+      {renderContent(gradientIdx)}
+
```
LucienLee, line 344: Need to add.
```
<defs>
  put your gradient definition here
</defs>
```

LucienLee (changes requested):

components/LevelIcon.js
MrOrz: Nice catch! `<defs>` is now added and UI does not break after add.
image

PR #354, opened 2020-11-12
Remove stackimpact

MrOrz:
Stack impact was introduced to mitigate CPU issue:
https://github.com/cofacts/rumors-site/pull/189

However after it's added, we did not get meaningful insight from it.
On the other hand, it requires native extension when installing the project and hence stops some of our contributor in g0v hackathon from contributing.

I suggest we remove stackimpact from rumors-site bundle.


nonumpa (approved):
LGTM!

bil4444 (approved):

LucienLee (approved):

Issue #265, opened 2020-06-23
Reply search result's time is not correct


MrOrz:
Reproduced on staging:
 
All replies are displayed as "replied less than 1 mins ago", but these replies should be created long time ago.
![螢幕快照 2020-06-23 下午9 51 23](https://user-images.githubusercontent.com/108608/85412519-3c55c100-b59c-11ea-9ff0-93ef85c4defd.png)

PR #355, opened 2020-11-23
fix reply time display bug

jojomango:
fix  https://github.com/cofacts/rumors-site/issues/265

createdAt is property of reply, not articleReply

### DIFF
--- a/components/ListPageDisplays/ReplyItem.js
+++ b/components/ListPageDisplays/ReplyItem.js
@@ -130,6 +130,7 @@ ReplyItem.fragments = {
         name
         ...AvatarData
       }
+      createdAt
       ...ArticleReplyFeedbackControlData
     }
     ${ArticleReplyFeedbackControl.fragments.ArticleReplyFeedbackControlData}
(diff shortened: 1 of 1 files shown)


MrOrz:
Hello @jojomango thanks for taking a look at the issue!

It turns out that `ArticleReply` actually have `createdAt` attribute on the API.
```graphql
"""The linkage between an Article and a Reply"""
type ArticleReply {
  replyId: String
  reply: Reply

  # ...(omitted)...

  status: ArticleReplyStatusEnum
  createdAt: String
  updatedAt: String
}
```

Seems that the root cause is that we forgot to include `articleReply`'s `createdAt` in our GraphQL queries. To be more specific, we forgot to include `createdAt` in [ReplyItemArticleReplyData fragment](https://github.com/cofacts/rumors-site/blob/3a135df69172617599e0dc7ebee704040a88fcf5/components/ListPageDisplays/ReplyItem.js#L125-L131). Would you take a look and give it a try? Thanks!

MrOrz (approved):
Fix confirmed on my localhost env. Thanks a million for taking a look at this issue and submit the concise and powerful fix!

PR #356, opened 2020-11-25
Landing page

hsiao19:
### Spec
figma: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=2760%3A463


components/AppLayout/AppLayout.js
```diff
@@ -97,6 +99,7 @@ function AppLayout({ children, container = true }) {
     };
     const handleRouteChangeComplete = () => {
       setRouteChanging(false);
+      window.scrollTo(0, 0);
```
MrOrz, line 102: I think the browser already handles scroll position across navigation.
Why should we reset scroll here?

components/AppLayout/AppLayout.js
```diff
@@ -111,14 +114,21 @@ function AppLayout({ children, container = true }) {
...
     <Fragment>
-      <AppHeader
-        user={data?.GetUser}
-        showProgress={isRouteChanging}
-        onMenuButtonClick={toggleSidebar}
-        onLoginModalOpen={openLoginModal}
-        onLogout={logout}
-        onNameChange={handleNameChange}
-      />
+      {pathname === '/' ? (
```
MrOrz, line 117: I prefer not using `AppLayout` in `src/index` at all, instead of adding more logic to `AppLayout` to support the landing page.

`AppLayout` is designed for application pages with header. It seems that the only common part between application pages and the landing page is the app footer and the login dialog. If that is the case, I suggest using `AppLayout/AppFooter` and `AppLayout/LoginModal` in `src/index` directly. Also, `LandingPageHeader` can be isolated from `AppLayout/AppHeader` and move to under `LandingPage` directory; since it's only used in `LandingPage`, it should be more straightforward to host it there.

MrOrz (commented):
Hello @hsiao19 thanks for the implementation!

After comparing the landing page with other pages in the application, I think it makes sense when we re-use `AppLayout/AppFooter.js` and `AppLayout/LoginModal.js`. Thank you for updating the style within these two components to the latest look!

However, I think we don't need to apply the entire `AppLayout/AppLayout.js`  in landing page (`src/index.js`), because landing page look and works differently from application content. I suggest that in `src/index.js`, we just import the components we want to re-use from `AppLayout` directory, but we do not need to wrap it with `<AppLayout>`, and `AppLayout` should not be modified to support the use in landing page.

MrOrz (commented):
Thanks for @hsiao19 for fixing key errors and the `InputBox` storybook & design!
JS review is completed, I am addressing CSS part in next review.

components/LandingPage/Stats.js
```diff
@@ -1,53 +1,219 @@
...
+    flexWrap: 'wrap',
+    maxWidth: 1134,
+    margin: '0 auto',
+
+    [theme.breakpoints.down('sm')]: {
+      maxWidth: 400,
+    },
+  },
+  item: {
+    display: 'flex',
```
MrOrz, line 82: Not turning `item` to flexbox container seems to make its contents align better.

### Current
image


### After removing `display: flex` and related styles
image

MrOrz (commented):
Sorry to keep you waiting. Here are the comments after reviewing the CSS.

hsiao19:
@MrOrz Both css and js have been revised, please check if there are any problems.

MrOrz (approved):
LGTM! Let's ship this :shipit:

PR #357, opened 2020-12-01
support openpeeps avatar

ztsai:
### storybook:
Screen Shot 2020-12-01 at 11 06 27 PM

### replies and feedbacks:
Screen Shot 2020-12-01 at 11 06 01 PM

### replyrequest:
Screen Shot 2020-12-01 at 11 05 09 PM

### reply list:
Screen Shot 2020-12-01 at 11 04 05 PM
Screen Shot 2020-12-01 at 11 03 54 PM

### header:
Screen Shot 2020-12-01 at 11 03 35 PM


components/AppLayout/Widgets/Avatar.js
```diff
@@ -71,23 +85,101 @@ const StatusBadge = withStyles(theme => ({
...
+          {...rest}
+        />
+      );
+    } catch {} // eslint-disable-line no-empty
+  }
+
+  if (!avatar) {
+    avatar = (
+      <img
+        className={cx(classes.root, className)}
```
MrOrz, line 175: is `mdSize` applied to this case as well?

components/ReplyRequestReason/ReplyRequestReason.js
```diff
@@ -18,16 +19,17 @@ const useStyles = makeStyles(theme => ({
...
-    fontSize: 40,
-  },
   vote: {
     borderRadius: 45,
     marginRight: 3,
     [theme.breakpoints.up('md')]: {
       marginRight: 10,
     },
   },
+  avatar: {
```
MrOrz, line 29: Could we drop `classes.avatar` as we specify `size={40}` in `<Avatar>`?

components/AppLayout/AppLayout.js
```diff
@@ -19,6 +19,9 @@ const USER_QUERY = gql`
       id
       name
       avatarUrl
+      avatarType
```
MrOrz, line 22: This actually comes from `Avatar` inside `AppHeader` and `AppSidebar`.

Since we follow [fragment colocation](https://www.apollographql.com/docs/react/data/fragments/#creating-colocated-fragments) in this repo, we may need to

- Declare static member `.fragments.User` for `AppSidebar` and `AppHeader` that includes `Avatar.fragments.AvatarData`, since these components are using `<Avatar>`
- AppLayout should use `AppSidebar.fragments.User` here. (apollo-client will perform dedup on the fields)

In this way the dependency of GraphQL fields can be tracked on component basis.

MrOrz (commented):
Hello @ztsai thanks for the contribution and sorry for the late reply. I have added comments regarding 1. use of `mdSize` and 2. colocating fragments.

Issue #285, opened 2020-07-11
“Add this reply to my reply” failed to copy reference


MrOrz:
Steps to reproduce
- Login
- Click "Reply to this message" in any article
- Choose "My reply" in search box
- Type in any keyword

After clicking “Add this reply to my reply” on searched result in editor, the contents of the copied reply appears in text area, but references shows "undefined".

![image](https://user-images.githubusercontent.com/108608/87227845-0acc5a80-c3d0-11ea-8348-213a6e2ac28d.png)

MrOrz:
Fixed by #358 . Kudos to @jojomango !

PR #358, opened 2020-12-06
fetch reference in reply

jojomango:
https://github.com/cofacts/rumors-site/issues/285

add reference in gql


MrOrz (approved):
Tested and works on my machine. Confirmed that this will fix #285. Thanks for the fix!

PR #365, opened 2020-12-19
Upgrade dialog

hsiao19:
Spec
https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=912%3A384


i18n/zh_TW.po
```diff
@@ -1234,4 +1234,35 @@ msgid ""
...
+#: components\AppLayout\UpgradeDialog.js:362
+msgctxt "upgrade dialog"
+msgid "I've got your back"
+msgstr "好的"
+
+#: components\AppLayout\UpgradeDialog.js:355
+#, javascript-format
+msgctxt "upgrade dialog"
+msgid "Cheers! ${ LEVEL_NAMES[nextLevel] }"
+msgstr "嗨嗨！闢謠天師"
```
MrOrz, line 1268: Please include variable `LEVEL_NAMES[nextLevel]` in the translation
```suggestion
msgstr "嗨嗨！${ LEVEL_NAMES[nextLevel] }"
```

MrOrz (changes requested):
Thanks for the implementation! I have some comments regarding the icons and state management. I will test on local machine after these logics got fixed.

MrOrz (commented):
Thanks for fixing the logic! Please update snapshot and take a look at possibility to simplify `repeating-conic-gradient` and we are good to go 👍

components/AppLayout/UpgradeDialog.js
```diff
@@ -15,95 +15,41 @@ import LEVEL_NAMES from 'constants/levelNames';
...
-const getConicGradient = (color1, color2, startAngle) => {
-  return `repeating-conic-gradient( 
-      from ${startAngle}deg, 
-      ${color1} ${startAngle}deg ${startAngle + 15}deg, 
-      ${color2} ${startAngle + 15}deg ${startAngle + 30}deg
-)`;
+const getSpinKeyframes = (steps, startDeg, color1, color2) => {
+  const keyframes = {};
+  const progressUnit = 100 / steps;
+  const degUnit = 360 / steps;
```
MrOrz, line 21: Since the pattern repeats every 30 degree and it is an repeating animation, we can just animate from 0deg to 30deg and have the same visual effect.

MrOrz (approved):
LGTM! The spinning logic is much simpler now :)

components/AppLayout/UpgradeDialog.js
MrOrz: Updated with `30deg / steps`.
With the same 10 steps, the animation became smoother because the increase of density:
spinner

MrOrz:
## Known issue

Due to limitation of Elasticsearch database, the user will need to submit one more reply for the level-up dialog to appear.

After submitting new reply & level up, the dialog will not be triggered until the user submits another reply in the same window.
The root cause is that by the time the apollo-client refetches `UserForUpgradeDialog` operation, Elasticsearch database is still indexing the new article-reply, thus the level number before inserting reply is returned instead.

User level queries are [aggregation queries](https://github.com/cofacts/rumors-api/blob/master/src/graphql/dataLoaders/userLevelLoaderFactory.js#L31-L61) that only updates after Elasticsearch index is refreshed.

## Proposed fix

We can either
- Refresh `articles` index inside [`createArticleReply`](https://github.com/cofacts/rumors-api/blob/c2ff8f2c8f7221bf431fb760438d09b01a8e6608/src/graphql/mutations/CreateArticleReply.js#L16), or
- Refresh `articles` index before user level query.

Both requires changes to data loader.

PR #367, opened 2020-12-19
[Profile-01] Profile page header display

MrOrz:
Implements user profile page's header.
Figma: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=912%3A384

- URL: `/user?id=xxx` or `/user/slug` (But no one has slug yet)
- Links to user page: on reply's header
- Edit function of profile page header is not implemented yet

### Clicking into user page
user-page-enter

### Self
(Edit button exists but its behavior is not implemented yet)

image

### Mobile
image


bil4444 (approved):

components/ProfilePage/ProfilePage.js
```diff
@@ -0,0 +1,84 @@
...
+`;
+
+function ProfilePage({ id, slug }) {
+  const currentUser = useCurrentUser();
+
+  const { data, loading } = useQuery(LOAD_USER, {
+    variables: { id, slug },
+  });
+
+  const isSelf = currentUser && data?.GetUser?.id === currentUser.id;
```
LucienLee, line 43: data?.GetUser?.id === currentUser?.id

LucienLee (approved):

components/ProfilePage/ProfilePage.js
johnsonliang-appier: Hmm if both `currentUser` and `data` is nullish (which is probably during the first render), `isSelf` will become true.

hsiao19 (approved):

PR #374, opened 2021-01-03
[Profile-02] Introduce <ProfileLink> component for links to user profile

MrOrz:
Implement `ProfileLink` to handle link to user profile.

User profile URL can be complex -- user with slug and user with ID has different URLs, hence different prop to `<Link>` tag. This PR:
- adds `ProfileLink` whose `fragments` loads necessary fields from `User` type to generate URLs precisely
    - Replace `EditorName` with more generalized `ProfileLink`
    - Replace occurrences of `EditorName` with `ProfileLink`
- `Avatar` now accepts `hasLink` prop.
    - When true, clicking the avatar brings visitor to user profile page
    - Make article reply feedback's avatar clickable

#### `hasTooltip` to & has slug
image

#### No tooltip, no slug
image


MrOrz:
Merged due to inactivity for >10 days

Issue #362, opened 2020-12-15
Provide better text for reference input


MrOrz:
Discussion see [20201209 meeting note](https://g0v.hackmd.io/@mrorz/cofacts-meeting-notes/%2FNqPf1SfPQEGOZ3Z6h7Q7wA)

## As-is
We just instruct editors to input 「資料來源」and provide「超連結與連結說明文字」in the placeholder.

<img width="901" alt="image" src="https://user-images.githubusercontent.com/108608/102174552-b8d8b500-3ed8-11eb-901a-3ca37c51578a.png">

### Title

This title should apply to website editors and the reply section below:
<img width="922" alt="截圖 2020-12-15 下午1 25 03" src="https://user-images.githubusercontent.com/108608/102174699-05bc8b80-3ed9-11eb-80f3-c2e49fea1e31.png">

#### When `replyType === 'OPINIONATED'`:
Opinion Sources
不同意見出處

#### When `replyType` is other values:
References
資料佐證

### Placeholder

One line summary
» Source URL

來源說明
» 連結網址

(Note: textarea placeholders [can include line breaks](https://stackoverflow.com/questions/7312623/insert-line-break-inside-placeholder-attribute-of-a-textarea). )

ulayab:
(舉手)

MrOrz:
Resolved by #381  @ulayab ++

MrOrz:
Reopen because we want to make placeholder 2 lines

```
One line summary
» Source URL
```

```
來源說明
» 連結網址
```

ulayab:
> Reopen because we want to make placeholder 2 lines
> 
> ```
> One line summary
> » Source URL
> ```
> 
> ```
> 來源說明
> » 連結網址
> ```

Thanks for the remind! Just make a new PR for it 🙏 https://github.com/cofacts/rumors-site/pull/382

MrOrz:
Closed in #382, kudos to @ulayab

MrOrz:
Reopen because the reply display should also align with To-be spec
![圖片](https://user-images.githubusercontent.com/108608/105885506-2df4e980-6044-11eb-8e57-c62d52c7e8e6.png)

PR #381, opened 2021-01-23
feat(ReferenceInput): #362 change text and translate

ulayab:
Using IIFE to workaround in line 61-66 because of unknown translate `ttag` issue.


MrOrz (approved):
Thanks for taking care of the translation and workaround the possible `ttag` bug! IIFE solution is very elegant and concise :)

PR #388, opened 2021-02-11
Implement tutorial page

hsiao19:
UI: https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=2011%3A0
Spec: https://g0v.hackmd.io/sB_zayWjTo-W0R7xe_U34w#%E5%88%86%E9%A0%85%E8%A7%A3%E8%AA%AA


.prettierrc
```diff
@@ -0,0 +1,28 @@
+{
```
MrOrz, line 1: Just curious:
Does eslint-plugin-prettier recognize this file?

MrOrz:
I am seeing that some image files are > 50KB. I suggest using [ImageOptim](https://imageoptim.com/mac) or pngquant (uses the same core as TinyPNG, but is easier to use since they run on local machine) and perform lossy compression on big image files first.

components/Slider.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    return () => {
+      if (slider) {
+        slider.removeEventListener('scroll', onScroll);
+      }
+    };
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [sliderRef, slidesAmount]);
```
MrOrz, line 127: I suggest we just add `initIndex` and `slideTo` to deps array and remove this hook. Adding these two dep should not break the whole thing, while disabling this may cause problem when we add code to this `useEffect` in the future.

```suggestion
  }, [sliderRef, slidesAmount, initIndex, slideTo]);
```

components/Slider.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
+    };
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [autoplay, activeSlide, interval]);
```
MrOrz, line 143: Same as above. It's [not safe](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) to omit functions from dep arrays.

As for `timeoutId`, it should really be a ref instead of a state. If it were a ref, we don't need to list it in dependency array either, as `ref.current` always points to the latest value, and we will never get a stale ref value coming from some unknown closure.

```suggestion
  }, [autoplay, activeSlide, interval, autoplayNext]);
```

MrOrz (changes requested):
Thanks for the PR. Happy new year!
I have added some suggestions regarding react-router `<Link>` and my thoughts on `Slider` props design.

Review progress:
image

components/Tutorial/Article.js
```diff
@@ -0,0 +1,335 @@
...
+  }, [isOpen, updateSubContentHeight]);
+
+  useEffect(() => {
+    setIsOpen(false);
+  }, [theme, title, subTitle, content, subContent]);
+
+  return (
+    <div className={cx(classes.root, className)}>
+      <div className={classes.label}>
+        <img src={theme === 'yellow' ? yellowLabel : blueLabel} />
```
MrOrz, line 253: Please consider using existing `<Ribbon>` instead of SVG. You can modify `<Ribbon>` to support blue background.

jest.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   moduleNameMapper: {
     // Ignore CSS and image imports in jest
     // Ref: https://jestjs.io/docs/en/webpack#handling-static-assets
-    '\\.(css|png|svg)$': '<rootDir>/__mocks__/styleMock.js',
+    '\\.css$': '<rootDir>/__mocks__/styleMock.js',
```
MrOrz, line 12: I think we can keep it simple, just use `fileMock` for the 4 file formats

pages/how-to-use.js
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <AppLayout container={false}>
+      <Head>
+        <title>{c('tutorial').t`how to use`}</title>
+      </Head>
+      <TutorialHeader />
+      <div className={classes.tabContainer}>
+        <Link href="/how-to-use?tab=bust-hoaxes">
+          <div
```
MrOrz, line 106: Suggest using `<a>` here so that the user will be able to cmd-click / long-press to open in new browser tabs, and it will also work for web crawlers to successfully follow the links

pages/how-to-use.js
```diff
@@ -0,0 +1,150 @@
...
+    query: { tab = 'bust-hoaxes' },
+  } = router;
+
+  return (
+    <AppLayout container={false}>
+      <Head>
+        <title>{c('tutorial').t`how to use`}</title>
+      </Head>
+      <TutorialHeader />
+      <div className={classes.tabContainer}>
```
MrOrz, line 104: Nit: we can use `nav` HTML tag so that web crawlers *may* recognized that it contains navigational links 😛

MrOrz (changes requested):
2nd batch of review comments. All previous commits have been reviewed.

MrOrz:
Thanks for the update! I have resolved comments that have been implemented. What's left are:
- `tutorial.js` image tab (`tabContainer`) should use `<a>`; consider using semantic tag `nav` for `tabContainer` 
- use `fileMock` and remove `styleMock`

MrOrz (approved):
LGTM! We will track items left in a separate ticket

PR #395, opened 2021-02-23
Edit avatar

ztsai:
continuing from https://github.com/cofacts/rumors-site/pull/357


components/AppLayout/Widgets/Avatar.js
```diff
@@ -72,24 +91,96 @@ const StatusBadge = withStyles(theme => ({
...
+          avatarData?.flip ? 'scale(-1, 1)' : 'scale(1, 1)'
+        } translateY(${size / 15}px)`,
+    },
+  },
+  // eslint-disable-next-line no-unused-vars
+}))(({ className, classes, avatarData, size, mdSize, ...rest }) => {
+  return (
+    <div className={className} {...rest}>
+      <div className={classes.showcaseWrapper}>
+        <Peep
```
MrOrz, line 129: This will create 3 levels of div around the `svg`.
The inner most div is created by `circleStyle` in `Peep` (relevant code: https://github.com/CeamKrier/react-peeps/blob/master/src/peeps/index.tsx#L160 )

I think we can
1. Just wrap 1 div around `<Peep>`
2. put styles in `peepsStyles.circleStyle` to `classes.showcaseWrapper`.
3. remove `circleStyle` prop from `<Peep>` to prevent `react-peeps` from generating extra div that

MrOrz (commented):
image
Some early comments so far

components/ProfilePage/AvatarSelector.js
```diff
@@ -0,0 +1,254 @@
...
+  const wrapperEl = useRef(null);
+  const classes = useStyles();
+
+  return (
+    <div ref={wrapperEl} className={classes.wrapper}>
+      {pieces.map(({ pieceType, type, viewBox, title }) => (
+        <ComponentInput
+          key={type}
+          pieceType={pieceType}
+          anchorEl={wrapperEl?.current}
```
MrOrz, line 231: When `ComponentInput` is rendered, `wrapperEl.current` is null; after `AvatarSelector` is rendered, `wrapperEl` is populated, but that does not trigger re-render.

`anchorEl` [supports a function](https://material-ui.com/api/menu/). I think when a function is passed, material-ui will invoke it when menu is opened, thus this can fetch the latest ref value in scope.

```suggestion
          anchorEl={() => wrapperEl.current}
```

Ref object returned by `useRef` is always an object that has `currrent` property (even just after initialization, it's `{current: null}`) so no need for `?.`.

The same applies to `<ColorPicker>` below.

components/ProfilePage/EditAvatarDialog.js
```diff
@@ -0,0 +1,203 @@
...
+              value="Github"
+              label="Github"
+              disabled={
+                !user.availableAvatarTypes ||
+                !user.availableAvatarTypes.some(val => val === 'Github') ||
+                !user.githubId
+              }
+            />
+          </Tabs>
+          <div className={classes.avatarEditor}>
```
MrOrz, line 169: Seems that removing this div can make mobile look even better; buttons can wrap and all buttons become clickable
image

And on desktop it looks decent as well.

components/ProfilePage/ProfilePage.js
```diff
@@ -84,6 +94,16 @@ function ProfilePage({ id, slug }) {
     }
   }, [latestSlug, router]);
 
+  // Load current-user-only fields when currentUser changes
+  useEffect(() => {
```
MrOrz, line 98: I think we can add [`refetchQueries: ['LoadProfilePage']`](https://www.apollographql.com/docs/react/data/mutations/#options) to `useMutation(UPDATE_USER)`  call in `EditAvatarDialog` instead. In this way, profile page data can be reloaded after user submits avatar.

components/ProfilePage/UserPageHeader.js
```diff
@@ -217,18 +307,26 @@ exported.fragments = {
       bio
       level
       appId
+
+      email
+      facebookId
+      githubId
+      availableAvatarTypes
```
MrOrz, line 314: It seems that these fields are only used in `EditAvatarDialog`. In this case I suggest adding these fields to `EditAvatarDialogUserData` fragment in `EditAvatarDialog` instead.

components/ProfilePage/UserPageHeader.js
```diff
@@ -217,18 +307,26 @@ exported.fragments = {
...
+
       points {
         total
         nextLevel
       }
       ...AvatarData
       ...LevelProgressBarData
       ...EditProfileDialogUserData
+      ...EditAvatarDialogUserData
```
MrOrz, line 323: Given the fact that
- All data in this fields are only required when EditAvatarDialog is opened; they are not required to render a profile page
- SSR for these fields won't work

I think it makes for sense if we perform a separate `useQuery` right inside render function of `EditProfileDialog` (i.e. after  `EditProfileDialog` is opened ) instead of putting its fragment in `UserPageHeader`'s `useQuery`. Plus, we could also drop `loadSelfOnlyFieldsForUser` lazy query & `useEffect` in `ProfilePage` if we load this fragment in `EditProfileDialog`.

(Seems that I should the similar to `EditProfileDialogUserData` as well; maybe I will refactor that in the future)

MrOrz (changes requested):
Thanks for the avatar edit function, it's a real fun to see these vibrant avatars!
I have some suggestions regarding data loading & DOM structuring, please take a look

components/ProfilePage/ProfilePage.js
ztsai: I updated user cached instead so it's updated in all avatars

components/AppLayout/Widgets/Avatar.js
```diff
@@ -124,15 +121,12 @@ const OpenPeepsAvatar = withStyles(theme => ({
...
-    <div className={className} {...rest}>
-      <div className={classes.showcaseWrapper}>
-        <Peep
-          {...omit(avatarData, ['backgroundColor'])}
-          style={peepsStyles.peepStyle}
-          circleStyle={peepsStyles.circleStyle}
-          strokeColor="#000"
-        />
-      </div>
+    <div className={`${className} ${classes.showcaseWrapper}`} {...rest}>
```
MrOrz, line 124: Use `cx` so that we won't end up with `undefined` in the `className` (which can be seen in snapshot)

```suggestion
    <div className={cx(className, classes.showcaseWrapper)} {...rest}>
```

MrOrz (approved):
It is looking great! Let me fix excessive `undefined` class name in snapshot, commit and we are ready to ship it :shipit:

PR #403, opened 2021-03-03
Impact page adjustment

hsiao19:
- wrap 'more' translation with ttag context
- handle influence background with media query


MrOrz (approved):
:shipit:

PR #422, opened 2021-04-03
[ProfilePage] fix slug infinite redirect

MrOrz:
Currently slugs with characters will trigger infinite page reload.

This PR modifies the reload detection mechanism so that it handles URL encoded paths correctly.


nonumpa (approved):
LGTM!

PR #432, opened 2021-05-02
Japanese version

MrOrz:
- Include Japanese translation in codebase
- Build & push script
- Relax landing page style so that it also works for Japanese version

Deployed to: https://ja.cofacts.tw/

localhost_3000_


bil4444 (approved):

MrOrz:
Seems that there are missing kanji font in our current font selection
image

Issue #417, opened 2021-03-17
Openpeeps avatar is not shown in landing page


MrOrz:
## As-is
![image](https://user-images.githubusercontent.com/108608/111423121-c5b0b300-872a-11eb-9650-7409f96834aa.png)

## To-be
Openpeeps avatar should show in the same way as in other pages 
![image](https://user-images.githubusercontent.com/108608/111423263-fbee3280-872a-11eb-9bac-27be22aa784e.png)

ulayab:
Hi @MrOrz
I just fixed this issue, and wondering which branch should I PR to ? thx🙏

MrOrz:
Hi @ulayab ,
Please open a PR to `master` branch. Thanks!

PR #438, opened 2021-07-04
fix: #417 landing page avatar displaying bug

ulayab:


MrOrz (approved):
Thanks a lot for finding the missing data!

I think it would be great if we can do the following to [achieve fragment colocation](https://www.apollographql.com/docs/react/data/fragments/#colocating-fragments):
1. move `USER_QUERY` to within `LandingPage/Header`
2. use [`Avatar.fragment.AvatarData`](https://github.com/cofacts/rumors-site/blob/9bece90df7d61e8fa298f3b3e99868e56bec698f/lib/useCurrentUser.js#L13) in query in the component that really uses `Avatar` component (that is `LandingPage/Header`)

@ulayab are you interested in changing the code above? I can do it after merge as well.

ulayab:
Sure! I'll try to deal with it.

ulayab:
Hi @MrOrz  I just update the branch.
BTW, it's my first time using Apollo, so I'm not really sure if it's an expected practice.🤧 
Please advise as necessary, thank you!

MrOrz (approved):
LGTM, thanks for the fix!

PR #442, opened 2021-08-08
Make ellipsis URLs copyable

MrOrz:
Currently when we copy links from "References" directly, we often copy the URLs with ellipsis in the middle, thus breaks the links.

This PR adjusted how ellipsis URLs are rendered so that when user selects the text and press "copy", the full URLs are copied instead.

Technically it is replacing `slice` + `...` with rendering the full URL + ellipsis with CSS.

## Can select & copy
Notice that the "..." in the URLs are gone when pasted.
copy-paste-full-urls

## Works in narrow screens
url-shrink


nonumpa (approved):
LGTM!

Issue #413, opened 2021-03-10
Contribution graph enhancements


MrOrz:
Ref: https://g0v.hackmd.io/@mrorz/cofacts-meeting-notes/%2F-Mth-vd5RPWqdLOfRJHHcg

- Color the contribution graph according to the user's max contribution count
- Crop old contribution like Github on mobile
- Use `Math.floor` to round-up the contribution to color

PR #453, opened 2021-10-06
add github like contribution graph css and ceil for 1,2 report count

kelvin2go:
https://github.com/cofacts/rumors-site/issues/413

- math.ceil for 1,2 
- add github contribution graph like css 

screen-capture (1)

### DIFF
--- a/components/ContributionChart.js
+++ b/components/ContributionChart.js
@@ -136,7 +136,9 @@ function Legend({ count }) {
 
 function scaleColor(count, maxContribution) {
   const varingScalingFactor =
-    count < 10 ? 10 / MAX_SCALE : maxContribution.count / MAX_SCALE;
+    maxContribution.count < 10
+      ? 10 / MAX_SCALE
+      : maxContribution.count / MAX_SCALE;
   return Math.max(
     Math.min(Math.ceil(count / varingScalingFactor), MAX_SCALE),
     0
--- a/components/__snapshots__/ContributionChart.stories.storyshot
+++ b/components/__snapshots__/ContributionChart.stories.storyshot
@@ -2907,7 +2907,7 @@ exports[`Storyshots ContributionChart Normal 1`] = `
                 >
                   <rect
                     aria-describedby={null}
-                    className="makeStyles-colorCofacts4"
+                    className="makeStyles-colorCofacts2"
                     height={10}
                     key="0"
                     onBlur={[Function]}
@@ -2985,7 +2985,7 @@ exports[`Storyshots ContributionChart Normal 1`] = `
                 >
                   <rect
                     aria-describedby={null}
-                    className="makeStyles-colorCofacts2"
+                    className="makeStyles-colorCofacts1"
                     height={10}
                     key="2"
                     onBlur={[Function]}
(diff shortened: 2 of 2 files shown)


kelvin2go:
```
adaptive max contribution
max(10, max(user_contribution)) 作為 contribution graph 的 max scale
```
for this, should it changes max scale to 10 ? 

https://github.com/cofacts/rumors-site/pull/453/files#diff-b170cff4fb01eac4d00fe240d828301e6e8900a7ae3443ad833eeff2a6d64d8fR13


and for test part should i update the story snapshot  after implement?

MrOrz:
Sorry for the late reply, and thanks for taking this issue!

- Current coloring of attribution is fixed at 5 contributions per level `SCALING_FACTOR`. Therefore, daily contribution must exceed 23 (`23 / SCALING_FACTOR = 4.6 ~= 5`) to be painted to deepest color.
- The *adaptive max contribution proposal* wants that max scale to adapt to each user. For example, if the max contribution count of the user is 16 for the past 365 days, we should color the contribution 13~16 with deepest color, and so on.
- However, if the max contribution count of the user for the past 365 days is lower than 10, it would be a bit unfair for other contributors if we also paint those contribution as deepest color. Therefore, in the proposal we added a max-contribution lower-bound: if the max contribution count of the user for the past 365 days is lower than 10, we will use 10 when calculating the color scale.

If we put it in code,
- `MAX_SCALE` remains 4, as available `colorCofactsN` still ranges from `colorCofacts0` to `colorCofacts4`.
- `SCALING_FACTOR` (number of contributions needed to go to the deeper color) should no longer be a constant across different users. If max contribution in `data` is below 10, the scaling factor is `10/MAX_SCALE = 2.5`; otherwise, it is `max contribution / MAX_SCALE`

nonumpa:
@kelvin2go It seems that your local material-ui version is different from this project.

You can try the command writing in [README.md](https://github.com/cofacts/rumors-site#storybook)

```
# install same npm package as this project
npm install

# run test and update snapshot
npm test -- -u
```

nonumpa:
It's abnormal, and I guess other .storyshot files also change `className="makeStyles-tooltipText-311"` to `className="makeStyles-tooltipText-74"`.

Maybe you can google the keyword "npm local vs global package" to solve this problem.

---
Explanation of why I guess your local material-ui version is different from this project and why it should not have `className="makeStyles-tooltipText-*"` change :

First, we take a look at the .storyshot change
snapshot

And then look what [CI test](https://github.com/cofacts/rumors-site/runs/3903371438) result says
> line 35 and 36 means
> It should be `aria-hidden="true"` not `aria-hidden={true}`

> line 48 and 49 means 
> It should be `className="makeStyles-tooltipText-311"` not `className="makeStyles-tooltipText-74"`

That means only `className="makeStyles-colorCofacts*"` is the right change.
This makes sense because you only change the function `scaleColor` which only affects `colorCofacts*`.

MrOrz:
No worries, I think I can make a commit that tries to fix CI

kelvin2go:
hi @MrOrz ,
after test with @nonumpa  
look like   "@material-ui/core": "^4.9.7", 

4.9.9 can pass the test.

not sure, which version is CI using @@'

PR #458, opened 2021-12-04
Handle SSG case

MrOrz:
[Previous image build](https://github.com/cofacts/rumors-site/actions/runs/1529029608) fails due to the server-side cookie logic introduced in #457 .

This PR fixes the build by properly handle SSG case, which will have undefined ctx.req.


MrOrz:
Build pass & works on staging, will directly merge

Issue #363, opened 2020-12-15
Fix editor RWD breakpoints


MrOrz:
From [20201209 meeting note](https://g0v.hackmd.io/NqPf1SfPQEGOZ3Z6h7Q7wA#mobile--desktop):

Breakpoint 設定有誤，mobile 與 desktop 版中間有一個出處沒有標題的版本：

![](https://s3-ap-northeast-1.amazonaws.com/g0v-hackmd-images/uploads/upload_7a650fa0654e0df850badd851fb0c489.png)

In this case please all go for desktop style.

Young-TW:
@MrOrz can close this issue?

MrOrz:
> @MrOrz can close this issue?
👌 👌 👌 
<img width="368" alt="圖片" src="https://user-images.githubusercontent.com/108608/145667829-926bf7a6-8ae1-4061-b904-537d5e9baba3.png">

PR #460, opened 2021-12-11
Fix editor RWD breakpoints #363

Young-TW:
edit breakpoints and css


MrOrz (approved):
Thanks a million for locating the root cause and fix the misaligned breakpoints! 🎉 

I found that the misaligned breakpoints also exist in `ReferenceInput` as well. 

## Desktop
圖片

## Tablet size (incorrect)
圖片

Also the border of textarea is missing.

With the information you give as you digging into this issue, I am able to fix the errors above real quick in a following PR. Kudos to @Young-TW for the analysis and the fix!

PR #484, opened 2022-04-28
Upgrade node and npm packages

MrOrz:
- Use `npm update` to Fix the dependabot PRs all at once
    - This updates the resolved minor version of all packages (including nextjs, material-ui, etc) in package-lock.json and somehow introduced many snapshot changes
- Update node version to 16
    - we cannot go 18 because webpack 4 (current storybook version) only supports nodejs 16 - https://github.com/webpack/webpack/issues/14532
    - Use `node:16-alpine` to shrink the production image size
- Upgrade storybook to 6
    - Fixes critical security holes reported by `npm audit`
    - Adds a storybook eslint plugin that suggests us on pascal case story naming convention
    - `main.js` supports `env` setup, which fixes the "process is not defined" error in statically built storybook
    - **`/storybook/index.html` will become viewable after this fix!**
- Upgrade jest to 26
    - We cannot go beyond jest 26 because jest >= 27 breaks storyshot (Ref: https://github.com/storybookjs/storybook/issues/15916 )
- Upgrade eslint


components/ActionMenu/__snapshots__/ActionMenu.stories.storyshot
```diff
@@ -33,7 +33,7 @@ exports[`Storyshots ActionMenu Default 1`] = `
         className="MuiButton-label"
       >
         <svg
-          aria-hidden="true"
+          aria-hidden={true}
           className="MuiSvgIcon-root"
           focusable="false"
```
nonumpa, line 38: Does storybook forget to change `focusable="false"` to `focusable={false}` ?

nonumpa (approved):
LGTM!

components/ActionMenu/__snapshots__/ActionMenu.stories.storyshot
MrOrz: Personally I don't know why this would change XD
For DOM elements, `{true}` and `"true"` are probably identical

PR #488, opened 2022-05-13
Don't show loading if article is already loaded

MrOrz:
Currently after up-voting or down-voting a reply, the whole article page will become "Loading...".

It seems that in @apollo/react-hooks v3.1.5 the `loading` state behavior is slightly different from 3.0.

This PR avoids "Loading..." page after an article is already displayed on screen.

upvote-downvote

### DIFF
--- a/pages/article/[id].js
+++ b/pages/article/[id].js
@@ -271,7 +271,7 @@ function ArticlePage() {
 
   usePushToDataLayer(!!article, { event: 'dataLoaded' });
 
-  if (loading) {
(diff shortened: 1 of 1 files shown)


MrOrz:
This is deployed to staging: https://dev.cofacts.tw/article/2889dua0k41ro

Everyone can test it out on staging article page above. We can deliver it to production on Saturday to see if anything breaks. This should help the editor's meetup on Sunday as well.

bil4444 (approved):

PR #513, opened 2022-10-29
Insert contact email in privacy policy sections

MrOrz:
Facebook says that the current status does not comply with platform policy, but it would after we do it this way.

Screenshot_20221029-162621


nonumpa (approved):
LGTM!

PR #528, opened 2023-02-11
adjust AppFooter

marcussfu:
截圖 2023-02-11 下午10 33 15


MrOrz (approved):
LGTM. Thanks for reusing the const defined in `constants/urls`!

PR #537, opened 2023-04-08
adjust 'Make editing comments (reply requests) easier'

marcussfu:
輸入comment後，按submit(送出)，輸入comment的text field會隱藏縮起來。
如果comments有自己之前輸入的comment，comment(我想補充)會變成update comment(更新補充)。
如果comments有自己之前輸入的comment，按update comment(更新補充)會把之前輸入的comment text先置入text field。

截圖 2023-04-08 下午4 26 19

### DIFF
--- a/components/CreateReplyRequestForm/CreateReplyRequestForm.js
+++ b/components/CreateReplyRequestForm/CreateReplyRequestForm.js
@@ -139,7 +139,13 @@ const SubmitButton = ({
 };
 
 const CreateReplyRequestForm = React.memo(
-  ({ articleId, articleUserId, requestedForReply, replyRequest, onNewReplyButtonClick }) => {
+  ({
+    articleId,
+    articleUserId,
+    requestedForReply,
+    replyRequest,
+    onNewReplyButtonClick,
+  }) => {
     const buttonRef = useRef(null);
     const [disabled, setDisabled] = useState(false);
     const [showForm, setShowForm] = useState(false);
@@ -162,7 +168,9 @@ const CreateReplyRequestForm = React.memo(
     }, []);
 
     useEffect(() => {
-      setText( replyRequest && replyRequest.length > 0? replyRequest[0].reason : text);
+      setText(
+        replyRequest && replyRequest.length > 0 ? replyRequest[0].reason : text
+      );
       setIsRequestedForReply(replyRequest && replyRequest.length > 0);
     }, [replyRequest]);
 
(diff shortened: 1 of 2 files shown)


components/CreateReplyRequestForm/CreateReplyRequestForm.js
```diff
@@ -139,11 +139,18 @@ const SubmitButton = ({
...
+    articleUserId,
+    requestedForReply,
+    replyRequest,
+    onNewReplyButtonClick,
+  }) => {
     const buttonRef = useRef(null);
     const [disabled, setDisabled] = useState(false);
     const [showForm, setShowForm] = useState(false);
     const [showFloatButton, setShowFloatButton] = useState(false);
+    const [isRequestedForReply, setIsRequestedForReply] = useState(false);
```
MrOrz, line 153: After testing on my local machine, I. can see that why you added `isRequestedForReply` -- it seems that `requestedForReply` boolean prop does not work properly. However, the current `useState` + `useEffect` implementation does not help displaying button text after the user refresh the page:
image

Since we have a new prop `replyRequest` now, I believe we can determine the value of `isRequestedForReply` directly from `replyRequest`, so the newly added `useState` and `useEffect` is probably not needed at all.

I would suggest
1. Remove `requestedForReply` prop from `CreateReplyRequestForm ` -- it's broken, and its purpose can be replaced by the new prop `replyRequest`
2. We either change `isRequestedForReply` to a boolean const that is calculated from `replyRequest`, or we just drop `isRequestedForReply` as well, because it's only used in one placed and its logic is simple enough to do it inline

pages/article/[id].js
```diff
@@ -364,6 +364,13 @@ function ArticlePage() {
       (currentUser ? true : positiveFeedbackCount - negativeFeedbackCount > 0)
   );
 
+  const replyRequestsWithCommentById = replyRequestsWithComments.filter(
```
Comment (MrOrz) [inline pages/article/[id].js:367]: The new property, `replyRequest`, is in singular form. I would suggest we also return the reply request itself, instead of returning an array.

We can use [`[].find`](https://developer.mozilla.org/zh-TW/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of filter to return the found element. It will return `null` if no such element is found.

```suggestion
  const ownReplyRequest = replyRequestsWithComments.find(
```

MrOrz (changes requested):
Thank you for looking into the issue and proposed a fix! I have some review comments and would like to see if you have time to fix them.

components/CreateReplyRequestForm/CreateReplyRequestForm.js
marcussfu: ok, i will try to fix.

pages/article/[id].js
Comment (marcussfu) [reply pages/article/[id].js:367]: ok, i will try to fix.

MrOrz (approved):
The solution now looks concise and good to read. Thanks for the contribution!

PR #543, opened 2023-07-11
Upgrade prettier, eslint and Typescript

MrOrz:
First step of modernizing rumors-site.
We will try updating Next.JS to at least 12 (which supports PostCSS 8 and should be able to install TailwindCSS) and then migrating to TailwindCSS if possible.


MrOrz:
Storybook build failed, will not be included in this week's release

MrOrz:
We will be implementing new site directly.

PR #556, opened 2024-01-06
adjust 'New feedbacks are not shown in reason popup when there are too many feedbacks'

marcussfu:
截圖 2024-01-06 下午4 03 52

adjust limit to 1000, and hide feedbacks that no comment.


MrOrz (commented):
Thanks for looking into the issue!

Back in the hackathon I thought hiding feedbacks without comments may help us focusing on those with comments. However, it comes to me that showing feedbacks without comment can help the audience in:
1. determine where the votes are from -- from other fact checkers, or from LINE users
2. better spot coordinated inauthentic behavior, when a group of fake accounts upvote each other and downvote other legitimate replies.

Therefore, I think we should _not_ filter out all those feedbacks that have no comments.

Instead, we can:
- Show those with comments first
- Then show comments without comments
- If there are more than 10 comments, put a "show all" link at the bottom of the list. Clicking them can show all the rest of the comments.

@marcussfu would you mind making the change?

marcussfu:
ok, I'll change it.

marcussfu:
Alread modified, check please.

MrOrz (approved):
It works very well, thank you for the change!

PR #559, opened 2024-01-10
fix(collaborate): blocked user shouldn't be able to edit transcript

nonumpa:
ref: https://g0v.hackmd.io/vKCvrqSQTlm7GEx9MAXGaw?view#op-Transcript-spam-%E8%99%95%E7%90%86

**normal user** 
截圖 2024-01-10 下午5 38 56
 
**blocked user**
截圖 2024-01-10 下午5 35 15


MrOrz (approved):
Tested on my machine, it works 👍🏻

Issue #274, opened 2020-07-11
Fold hyperlink previews in detail pages


MrOrz:
Implement hyperlink folding when there are more than 4 hyperlinks. [Visual spec](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=2079%3A939)

![image](https://user-images.githubusercontent.com/108608/87218122-81426b80-c382-11ea-9f4b-d8852d09f385.png)

- 1~3 hyperlinks: show all
- 4+ hyperlinks: fold to 2 hyperlinks and a "show" button
- No "collapse" button after expansion

jiru:
Working on this.

PR #571, opened 2024-07-20
Enhance display of many link previews

jiru:
Solves #274.

Changes:
* When on mobile, and when then are two or more links, the links are displayed horizontally and can be scrolled horizontally.


MrOrz (approved):
Thank you for the contribution! The new hyperlinks looks beautiful on mobile.

PR #581, opened 2024-11-23
fix: feedbacks aren't sorted as expected

belong112:
截圖 2024-11-23 下午3 15 41

feedbacks with comments are sorted in the front

### DIFF
--- a/components/ArticleReplyFeedbackControl/ReasonsDisplay.js
+++ b/components/ArticleReplyFeedbackControl/ReasonsDisplay.js
@@ -53,6 +53,7 @@ export const LOAD_FEEDBACKS = gql`
         node {
           id
           vote
+          createdAt
           user {
             id
           }
@@ -67,6 +68,21 @@ export const LOAD_FEEDBACKS = gql`
   }
   ${Feedback.fragments.ReasonDisplayFeedbackData}
 `;
+
+function isEmptyComment(comment) {
+  return comment === '' || comment === null;
+}
+
+function processedFeedbacks(feedbacks, voteType, isLoadMore) {
+  return feedbacks
+    .filter(({ vote }) => vote === voteType)
+    .sort(
+      (a, b) =>
+        isEmptyComment(b.comment) - isEmptyComment(a.comment) ||
+        b.createdAt.localeCompare(a.createdAt)
+    )
+    .slice(0, isLoadMore ? feedbacks.length : Math.min(feedbacks.length, 10));
+}
 
 function ReasonsDisplay({ articleReply, onSizeChange = () => {} }) {
   const classes = useStyles();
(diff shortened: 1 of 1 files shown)


components/ArticleReplyFeedbackControl/ReasonsDisplay.js
```diff
@@ -95,9 +95,21 @@ function ReasonsDisplay({ articleReply, onSizeChange = () => {} }) {
     if (onSizeChange) return onSizeChange();
   }, [tab, onSizeChange]);
 
+  const IsEmptyComment = comment => comment === '' || comment === null;
```
MrOrz, line 98: nit: please help moving this helper function to outside of the component, so that the browser don't create this function on every render.
Also, we can use camelCase instead of PascalCase, which is mostly reserved for React components.

components/ArticleReplyFeedbackControl/ReasonsDisplay.js
```diff
@@ -95,9 +95,21 @@ function ReasonsDisplay({ articleReply, onSizeChange = () => {} }) {
...
+
   const feedbacks =
     data?.ListArticleReplyFeedbacks?.edges.map(({ node }) => node) || [];
 
+  const processedFeedbacks = (feedbacks, voteType, isLoadMore) =>
```
MrOrz, line 103: nit: we can also move this processor function to outside of the component to avoid re-creating the function on everyrender.

components/ArticleReplyFeedbackControl/ReasonsDisplay.js
```diff
@@ -95,9 +95,21 @@ function ReasonsDisplay({ articleReply, onSizeChange = () => {} }) {
...
+    feedbacks
+      .filter(({ vote }) => vote === voteType)
+      .sort((a, b) => {
+        if (IsEmptyComment(a.comment)) return 1;
+        else if (IsEmptyComment(b.comment)) return -1;
+        return 0;
```
MrOrz, line 109: nit: I am thinking if we should sort feedbacks using time (can use [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) on `.createdAt` or `.updatedAt) when both `a` and `b` are populated (or when both are not populated)

MrOrz (commented):
Hello, thank you so much for looking into this issue! I have some comments on the PR. Would you make further changes on the PR, or would you prefer me to do the changes instead?

belong112:
> Hello, thank you so much for looking into this issue! I have some comments on the PR. Would you make further changes on the PR, or would you prefer me to do the changes instead?

Sure, thanks for reviewing and giving comments !

components/ArticleReplyFeedbackControl/ReasonsDisplay.js
belong112: I've used createdAt as sorted key, the newest feedbacks will go first

belong112:
Modified, please check~

MrOrz (approved):

Issue #563, opened 2024-01-17
Redirect to new user profile URL after the user removes their slug settings


MrOrz:
From 20240110 test result https://g0v.hackmd.io/tiTBOmf_TaGYL9THZtQ9tg#%E6%9C%AA%E7%AB%9F%E9%A0%85%E7%9B%AE1

remove slug 後 user link 不會更新，按 F5 會壞掉 [name=cai]
  - 應該也要 redirect 到 ?id=xxx 的 page 避免此狀況
  - 但這比較少發生，可以另外開票做 [name=mrorz]

belong112:
taking this issue~

MrOrz:
During test, we found that "This user does not exist" still happens, but at least it's less likely to happen.
https://g0v.hackmd.io/@cofacts/meetings/%2FJBzxtPVYTbattxTcYTVRgA#%E6%9C%AA%E7%AB%9F%E9%A0%85%E7%9B%AE

belong112:
Yeah, I think is the problem of waiting for mutation. This also happens before my change.

PR #582, opened 2024-11-25
fix: router redirect when no id and no slug

belong112:
When entering user page without id and without slug (`/user/`), it will redirect to `/user?id={userId}`

### DIFF
--- a/components/ProfilePage/ProfilePage.js
+++ b/components/ProfilePage/ProfilePage.js
@@ -110,6 +110,12 @@ function ProfilePage({ id, slug }) {
   }
 
   if (!data || !data.GetUser) {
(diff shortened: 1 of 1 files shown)


MrOrz (commented):
Thanks for the contribution. I would like to know some background on this fix.

1. the reason why do we want to support `/user` redirecting to current user
2. would `router.replace` affect server-side rendering?

belong112:
Oops, I misunderstood issue #563. I will close this PR and rethink the solution.

MrOrz:
Ohh I see that you want to tackle with #563. How about performing the redirect when we finish modification of slug?

belong112:
No problem, I will start a new PR.

PR #583, opened 2024-11-25
fix: feedbacks were not sorted correctly

belong112:
feedbacks with comment should be sorted to the front
螢幕擷取畫面 2024-11-25 153255


MrOrz (approved):
Thanks for the fix! Let me deploy it to staging and take a look

PR #585, opened 2024-11-26
fix(ProfilePage): redirect to new user profile URL after user changin…

belong112:
Did a few adjustments
1. Changed the condition `!latestSlug` to `latestSlug === undefined` since slug might be cleared.
2. Redirect to `/user?id={yourId}` when `latestSlug` is null or empty.
3. Replaced `window.location.pathname` with `router.asPath` because `window.location.pathname` cannot capture query parameters.


MrOrz (approved):
The logic is crystal clear. Let me merge it and try it on staging. Thank you for the contribution!

MrOrz:
Got a new issue, not sure if its related to this PR: https://github.com/cofacts/rumors-site/issues/587

Issue #412, opened 2021-03-10
Show join date on Profile page


MrOrz:
See discussion in [20210303](https://g0v.hackmd.io/-Mth-vd5RPWqdLOfRJHHcg#Show-join-date-on-profile-page)

We should display join date of a user on their profile page.

MrOrz:
Completed by #586

PR #586, opened 2024-12-17
feat(Profile page): show join date

belong112:
Display join date of a user on their profile page #412 

In English
截圖 2024-12-18 上午12 26 06

In Mandarin
截圖 2024-12-18 上午12 27 54

Not sure whether putting JoinDate in that place is optimal for UI


MrOrz (approved):
Looks good to me, thank you for the contribution!

As for the placement, I think we can merge now and make adjustments if needed

PR #588, opened 2024-12-20
fix(ProfilePage): should see other users' profile page

belong112:
## explaination
I found that latestSlug returns null when the user hasn't set their slug yet. Therefore, I changed `latestSlug === undefined` to `!latestSlug && latestSlug !== ''`
Additionally, if the user resets their slug to an empty string, it will fail the condition. To address this, I changed the userId to ensure redirection leads to the user's page instead of "my" page.

## preview
截圖 2024-12-21 上午1 59 54
able to see other user's profile page


MrOrz (approved):
LGTM! Let's try it on staging now

MrOrz:
We found that the comments tab under user profile page is broken, not sure if it is related: #594

PR #589, opened 2024-12-21
feat: send upvote / downvote without reason

### LINKED ISSUE #294: Send upvote / downvote at the same time button is pressed
From: https://g0v.hackmd.io/OC7BneJ3TEGJHge1hWV-0w?both#%E6%9C%AA%E7%AB%9F%E9%A0%85%E7%9B%AE

## As-is
image

- Clicking upvote / downvote shows reason dialog.
- Upvote / downvote are sent along with its reason after clicking "Send".

## To-br
- Upvote / downvote (without reason) is submitted when button is clicked.
- Clicking upvote / downvote still shows reason dialog, but with different prompt: "Thank you for the feedback." + "Do you have anything to add?" / "Why do you think it is not useful?"
- "Send" --> "Close" when no content, "Send" when inputted something
- After clicking "Send", the vote is submitted again, this time with reason.

ydcjeff:
fix #294

As in the issue description, I have changed to submit the feedback on every upvote / downvote click. The reason dialog stays the same, and it shows `Close` when there is no input text, `Send` otherwise.

### DIFF
--- a/components/ArticleReplyFeedbackControl/ArticleReplyFeedbackControl.js
+++ b/components/ArticleReplyFeedbackControl/ArticleReplyFeedbackControl.js
@@ -136,6 +136,13 @@ function ArticleReplyFeedbackControl({ articleReply, className }) {
   const openVotePopover = (event, value) => {
     setVotePopoverAnchorEl(event.currentTarget);
     setVote(value);
+    createReplyFeedback({
+      variables: {
+        articleId: articleReply.articleId,
+        replyId: articleReply.replyId,
+        vote: value,
+      },
+    });
   };
 
   const closeVotePopover = () => {
@@ -228,7 +235,7 @@ function ArticleReplyFeedbackControl({ articleReply, className }) {
               });
             }}
           >
-            {t`Send`}
+            {reason ? t`Send` : t`Close`}
           </Button>
         </div>
       </Popover>


components/ArticleReplyFeedbackControl/ArticleReplyFeedbackControl.js
```diff
@@ -136,6 +136,13 @@ function ArticleReplyFeedbackControl({ articleReply, className }) {
   const openVotePopover = (event, value) => {
     setVotePopoverAnchorEl(event.currentTarget);
     setVote(value);
+    createReplyFeedback({
+      variables: {
+        articleId: articleReply.articleId,
+        replyId: articleReply.replyId,
+        vote: value,
```
MrOrz, line 143: This will cause issue in the following user sequence:
1. The user clicks upvote, fills in some reason and then send
2. The user clicks upvote again to open popup
3. The user tries to dismiss the popup

We expect that the user's reason in step 1 should be preserved. However, the current implementation will wipe the reason entirely.

We may argue that the existing feedback button does not load the reason either. However, in this PR, it makes it easier for the user to clear the reason by accident.

I suggest we try amending the reason loading logic before accepting this change.

ydcjeff: So, if I understand correctly, we want to preserve the user submitted reason when they open the popup?
And optionally create a reply feedback on upvote / downvote with the existing reason if it exists?

MrOrz: Yes -- when we submit feedback upon popup is opened, it would be great if the reason is preserved. If the user want to change their reason after popup is opened, it's easier for them when the popup is already populated with existing reason.

PR #590, opened 2024-12-21
feat: auto resize search bar

### LINKED ISSUE #281: Search bar input should span multiple rows
## As-is
The search input in search page is only 1 line high
search

## To-be
As [in spec](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO/Cofacts-website?node-id=432%3A2727), the search input should grow as input.

image

This should be achievable using Material UI's [Textarea autosize](https://material-ui.com/components/textarea-autosize/).

ydcjeff:
fix #281 

As per requirements from the issue description, I have used `TextAreaAutoResize` component from Material UI. I have also adjusted the styles of the textarea since the text inside is overlapping with search icon button.

Before:
image

After:
image

I noticed pressing enter key on the current site https://cofacts.org/search did not trigger search, but I have added it as per the issue description.

I also adjusted the `Searching` text to stick at the top of the flex area.

### DIFF
--- a/components/SearchPageJumbotron.js
+++ b/components/SearchPageJumbotron.js
@@ -116,7 +116,7 @@ function SearchPageJumbotron() {
               name="search"
               className={classes.input}
               minRows={1}
-              onKeyDown={(e) => {
+              onKeyDown={e => {
                 // enter to trigger search, so disable default line breaking
                 if (e.key == 'Enter') {
                   e.preventDefault();
--- a/components/SearchPageJumbotron.js
+++ b/components/SearchPageJumbotron.js
@@ -2,7 +2,7 @@ import { useRef, useEffect, memo } from 'react';
 import { useRouter } from 'next/router';
 import { t } from 'ttag';
 import { makeStyles, withStyles } from '@material-ui/core/styles';
-import { Tabs, Tab, Box, Container } from '@material-ui/core';
+import { Tabs, Tab, Box, Container, TextareaAutosize } from '@material-ui/core';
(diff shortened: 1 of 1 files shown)


components/SearchPageJumbotron.js
```diff
@@ -109,11 +111,18 @@ function SearchPageJumbotron() {
...
-            <textarea
+            <TextareaAutosize
               ref={textareaRef}
               name="search"
               className={classes.input}
-              rows={1}
+              minRows={1}
+              onKeyDown={e => {
+                // enter to trigger search, so disable default line breaking
+                if (e.key == 'Enter') {
```
MrOrz, line 121: I am thinking if we should allow line break when shift-enter is pressed. What do you think?

MrOrz (commented):
Thanks for your contribution! I have raised a UX-related question

components/SearchPageJumbotron.js
ydcjeff: I haven't looked into how search works, but given that the use case is searching, I personally think that users will input space separated texts instead of line separated texts.

MrOrz: Agree. it's just that when user are provided with a multiline text area that submits when hitting enter, they may expect a soft-return mechanism (usually shift+enter) so that they can enter new lines. An example would be message textarea of a chat window.

But this is not mandatory for Cofacts' search box; I would expect that most user would copy and paste suspicous text from elsewhere and directly hit enter.

PR #591, opened 2024-12-21
feat: show login modal when anonymous user clicks on comment btn

### LINKED ISSUE #277: Block anonymous user from clicking "Reply to this message" and "Comment" button
#### Reply to this message
Expands an empty section
Jul-11-2020 18-54-42

#### Comment section
Allows user to input long comments, but block on submission
comment-click

### Mobile
Block the user from doing anything......
mobile

## To-be
[Designers said](https://www.figma.com/file/zpD45j8nqDB2XfA6m2QskO?node-id=2011:0#30609561):
image

Let's keep the buttons clickable, but add an `alert()` asking the user to login immediately after it's clicked.
(Popping up login screen can be even better, but it's a bit more difficult to achieve.)

ydcjeff:
fix #277

This PR changes to show the login modal when the user is not logged in and the user wants to comment on an article.

I have done this with React Context instead of prop drilling.
I am not quite familiar with React, please let me know it there are any changes needed.

### DIFF
--- a/components/AppLayout/AppLayout.js
+++ b/components/AppLayout/AppLayout.js
@@ -13,6 +13,7 @@ import LoginModal from './LoginModal';
 import fetchAPI from 'lib/fetchAPI';
 import { blockUserBrowserAndRefreshIfNeeded } from 'lib/isUserBlocked';
 import Snackbar from '@material-ui/core/Snackbar';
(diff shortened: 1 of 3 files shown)


MrOrz (approved):
Thanks for the fix, let's try it out on staging

PR #592, opened 2024-12-21
feat: add hash to reply in article detail page

### LINKED ISSUE #348: Add hash to new reply section & each reply in article detail page
From slack discussion:

image

Related material:
- https://www.bram.us/2020/03/01/prevent-content-from-being-hidden-underneath-a-fixed-header-by-using-scroll-margin-top/
- https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

ydcjeff:
fix #348 

This PR adds a hash symbol after each reply summary, scroll into view on click or on page load.

### DIFF
--- a/components/ArticleReplySummary.js
+++ b/components/ArticleReplySummary.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import { t, jt } from 'ttag';
 import gql from 'graphql-tag';
 import { ProfileTooltip } from 'components/ProfileLink';
@@ -25,11 +25,16 @@ const useStyles = makeStyles(theme => ({
     '& > a': { textDecoration: 'none' },
     '& > a:hover': { textDecoration: 'underline' },
   },
+  hash: {
+    marginLeft: theme.spacing(1),
+    scrollMargin: `${theme.spacing(15)}px`,
+  },
 }));
 
 function ArticleReplySummary({ articleReply, className, ...props }) {
   const { replyType, user } = articleReply;
 
+  const anchorRef = useRef();
   const classes = useStyles({ replyType });
 
   const authorElem = (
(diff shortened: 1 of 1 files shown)


components/ArticleReplySummary.js
```diff
@@ -25,11 +25,16 @@ const useStyles = makeStyles(theme => ({
     '& > a': { textDecoration: 'none' },
     '& > a:hover': { textDecoration: 'underline' },
   },
+  hash: {
+    marginLeft: theme.spacing(1),
+    scrollMargin: `${theme.spacing(15)}px`,
```
ydcjeff, line 30: I chose this value to look good on both desktop and mobile.

components/ArticleReplySummary.js
```diff
@@ -38,9 +43,23 @@ function ArticleReplySummary({ articleReply, className, ...props }) {
     </ProfileTooltip>
   );
 
+  useEffect(() => {
+    if (anchorRef.current.hash === location.hash) {
+      anchorRef.current.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, []);
```
MrOrz, line 50: Personally I prefer [scroll-behavior CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) than imperative JS scrolling. I think it's OK to apply scroll-behavior globally

components/ArticleReplySummary.js
```diff
@@ -38,9 +43,23 @@ function ArticleReplySummary({ articleReply, className, ...props }) {
...
+
   return (
     <div className={cx(classes.replyType, className)} {...props}>
       {jt`${authorElem} mark this message ${TYPE_NAME[replyType]}`}
+      <a
+        ref={anchorRef}
+        className={classes.hash}
+        href={`#${articleReply.replyId}`}
+        onClick={e => e.target.scrollIntoView({ behavior: 'smooth' })}
```
MrOrz, line 59: `scrollIntoView` is probably not needed if we define a global scroll-behavior in CSS.

However, it would be nice if clicking `#` copies the URL + href into clipboard. We are already using `copy-to-clipboard` package.

